PR #76, opened 2015-10-30
fix DynamicAnalyzer post 500 status

xl7dev:


ajinabraham:
@xl7dev : The reason for 500 status might be something else. What made you to this conclusion?
Shell=True what does it fix here?

xl7dev:
Access http://127.0.0.1:8000/DynamicAnalyzer 500 error, I tried command chmod 777 DynamicAnalyzer/tools，but the sames error,access python website https://docs.python.org/2/library/subprocess.html#subprocess.call fix the bug

PR #170, opened 2016-06-02
Fixes #126. Replaced codecs with io module.

### LINKED ISSUE #126: API Fuzzer: bug replace codecs with io
https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/blob/55f42771b1992e1d27519bd6a431ddf11c81eb0c/DynamicAnalyzer/pyWebProxy/pywebproxy.py#L88

The problem is due to a bug in the "codec" library that detects the character
"F4" and assumes this is the first character of a sequence of characters and wait
to receive the remaining 3 bytes, and the resulting string is truncated.
A better and safer approach would be to read the entire stream and only then
proceed to the decoding phase, as done by the ”io” module.

ref: https://www.owasp.org/images/c/cb/OWASP_Romania_Branca.pdf
page 29

c0dist:
Not much efforts was required, since codecs and io have kind of same definition for "open".

### DIFF
--- a/DynamicAnalyzer/pyWebProxy/pywebproxy.py
+++ b/DynamicAnalyzer/pyWebProxy/pywebproxy.py
@@ -47,7 +47,7 @@ import ssl
 import os
 import datetime
 import uuid
-import re,sys,threading,json,codecs
+import re,sys,threading,json,io
 from multiprocessing import Process, Value, Lock
 from socket_wrapper import wrap_socket
 from MobSF.utils import PrintException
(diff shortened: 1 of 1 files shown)


ajinabraham:
This needs to be tested thoroughly. Else where in MobSF we use `io.open`.
If I remember correctly `codecs.open` was used for the proxy due to some unicode error with `io.open`
That's the reason I didn't fixed it at first place. I will do more testing with `io` module on APIFuzzer. If everything goes good, I will merge this.

c0dist:
Sure thing. This sounds like a good plan. Since I don't have a running instance of MobSF, I cannot test it out. However, if you have the data causing error with `io` module, I will be able to test it out for you.

ajinabraham:
Migrated to `pickle`. `io` or `codec` modules are not required anymore.

PR #224, opened 2016-09-06
Basic windows phone app static analysis

DominikSchlecht:
### What has been implemented
- Static analysis of windows phone apps (currently only appx)
    - Parsing the xml-files
    - Static analysis with binskim and binscope
- Sadly a windows vm is needed, but a setup-script is located in install/windows/setup.py and the install/windows/readme.md gives a step by step tutorial
- Tested with 4 different apps; Setup tested on a blank windows 10 64x

### What has to be done before pushing
- Correct download-paths in install/windows/setup.py, install/windows/readme.md and install/windows/config.txt (I left the download path to my repository so you could test the code without pushing the files to your repo)

### What's to come
- Support for other windows phone file formats
- Deeper analysis
- Moving the windows vm to a docker image (don't know yet if this will work out and there are quite some showstoppers right now)

If you have questions or if I should rework some files, please feel free to contact me here or via e-mail.

Greetings,
Dominik


ajinabraham:
@DominikSchlecht great Work!
I will do a code review, manual integration test and will let you know about it.

Regarding `Sadly a windows vm is needed`
I suppose the Windows VM is required because the binary analysis tools are PE and require Windows to run.
Can we have a configurable option like if the host OS is Mac or Linux, a Windows VM is required and if the Host OS is Windows itself, can't we run the binary analysis tools in Host Windows environment itself? I assume that during static analysis, you are not running any executables from the appx package. So it will help a user to reduce the configurations needed.

DominikSchlecht:
@ajinabraham Thanks for the great framework :)

Yes, Binskim and Binscope are windows-applications, that's why a windows machine is required. You are right, I will include a feature that checks for the currently running platform and will allow for local analysis if the user is on windows 👍 Thanks for the Input.

ajinabraham:
Moved to a new branch so that I can test things effectively. https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/pull/225

Added some minor changes: https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/pull/225/commits/728e6d5a758ba4bbf40901994b3db645ea0544dc
Have a Look.

Code Review of the Framework related code is done. Looks Good to me

Few things to do.
* `strings` is supported only in Unix based OS so embed strings.exe for windows. (Place it under tools/windows dir in static analyzer
* A copy of Binskim and Binscope if they are portable under tools/windows (Analysis to do in Windows Host)
* Supports PDF report generation
* Some kind of small auth to access the API in Flask RPC server as it's exposed to `0.0.0.0`. Host only is a good option but we cannot always assume that people will go by that way. 
* `upload_file` in Flask Server should have proper MIME Type checks

ajinabraham:
Sent you an invite as a collaborator. Now you can add any changes/updates regarding Windows Static Analysis to `windows_static` branch

DominikSchlecht:
@ajinabraham thanks, will try to fix the things in the upcoming week 👍

DominikSchlecht:
Dear @ajinabraham,
some considerations:
- `strings is supported only in Unix based OS so embed strings.exe for windows. (Place it under tools/windows dir in static analyzer`
    Sadly the eula for the strings.exe states:
> You may not
>    [...]
>        * publish the software for others to copy;
>    [...]

    Since I do this partly in my work-time for my employer, I can not risk to break software agreements, sorry :/

- `A copy of Binskim and Binscope if they are portable under tools/windows (Analysis to do in Windows Host)`:
    Sadly for Binscope there is only an installer and if the binary is moved, it doesn't work anymore (I have no idea why someone would write a tool like this but ok..).
    Binskim seems to be the only tools I could really add to MobSF, but it would be about 25mb. Are you sure that I should include the files? Also, there might be a better way (see below).

Since it seems like a bad idea to include Microsoft software directly, would it be ok for you if I write another little installer script like the one for the VM for windows host systems? It would be triggered when MobSF is run for the first time and windows is detected as the host system. This would save space for everyone not using windows as the host system and the user would always have the newest version.

ajinabraham:
@DominikSchlecht 
* Regarding Strings, we need to see how we can use this or similar utilities alternatively.
https://gist.github.com/berdario/114b2daf9b43fe924676
http://stackoverflow.com/questions/17195924/python-equivalent-of-unix-strings-utility

*  We can keep an install instructions for those two binaries or like you said an install script or a batch file. But going forward if the windows binaries that we need to use are portable and permitted by licence, and have smaller size,  we should think of bundling them to make the process easier.

ajinabraham:
Closing this PR as the development of Windows Static Analyzer is tracked in `windows_static` branch and this PR: https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/pull/225

PR #231, opened 2016-09-29
Implemented features/bugfixes for windows static analysis

DominikSchlecht:
Implemented/Fixed:
- Multi-OS-String-Command via Python (d8b0a5f16f64c1fbf7a8a8ed6532f21d2f7aa9df)
- PDF Report (1220781a3396ea693e2542d8062ca81c89233062)
- Changed from Flask to XMLRPC (0f8077299be6675315cfc513f50a0ce72b0ba2ed, 3de3460c64876d0432a3690dba985a5fa18df01a)
- Auth for XMLRPC (0f8077299be6675315cfc513f50a0ce72b0ba2ed, 3de3460c64876d0432a3690dba985a5fa18df01a)
- Added MIME-Types (c03216ec48370bd020fd721a4a71259538522af6)
- Ability to run the analysis on windows without an extra VM
    - setup.py is now callable from python3 and python2 (608c14089a6206cef4e46c9a7c749f101b3089bd)
    - settings.py checks for os and runs the installer (only needed methods) one time (28fcf826c0f056a9906c772a08f7f22a9feeea68)
    - binary_analsis in windows.py checks settings.py for os and executes locally or via rpc (28fcf826c0f056a9906c772a08f7f22a9feeea68, 8aae3453ee1a0bef51990010ea3e14777b715d92)


StaticAnalyzer/tools/strings.py
```diff
@@ -0,0 +1,21 @@
...
+
+def strings(filename, min=4):
+    """Print out all connected series of readable chars longer than min."""
+    with open(filename, "rb") as f:
+        result = ""
+        for c in f.read():
+            if c in string.printable:
+                result += c
+                continue
+            if len(result) >= min:
```
ajinabraham: We probably need to think on the `min` also checks like result should start with a number or alphabet and not a special char.

ajinabraham: Sample Code:

```
if len(result) >= min and result[0].isalnum():

```

StaticAnalyzer/views/shared_func.py
```diff
@@ -187,6 +190,37 @@ def PDF(request):
...
+                            'compiler_version' :  db_entry[0].COMPILER_VERSION,
+                            'visual_studio_version' :  db_entry[0].VISUAL_STUDIO_VERSION,
+                            'visual_studio_edition' :  db_entry[0].VISUAL_STUDIO_EDITION,
+                            'target_os' :  db_entry[0].TARGET_OS,
+                            'appx_dll_version' :  db_entry[0].APPX_DLL_VERSION,
+                            'proj_guid' :  db_entry[0].PROJ_GUID,
+                            'opti_tool' :  db_entry[0].OPTI_TOOL,
+                            'target_run' :  db_entry[0].TARGET_RUN,
+                            'strings' : db_entry[0].STRINGS,
+                            'bin_an_results' : ast.literal_eval(db_entry[0].BIN_AN_RESULTS),
```
ajinabraham: Wouldn't it be good to use the existing util functions for this:
https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/blob/master/MobSF/utils.py#L238-L245

StaticAnalyzer/views/windows.py
```diff
@@ -23,8 +28,16 @@
 
 from StaticAnalyzer.models import StaticAnalyzerWindows
 
+from StaticAnalyzer.tools.strings import strings
+
 from MobSF.utils import PrintException
 
+
+proxy = xmlrpclib.ServerProxy( # pylint: disable-msg=C0103
+    "http://{}:8000".format(settings.WINDOWS_VM_IP)
```
ajinabraham: Can we make port configurable and move it to settings.py

StaticAnalyzer/views/windows.py
```diff
@@ -177,6 +192,14 @@ def staticanalyzer_windows(request):
         template = "error.html"
         return render(request, template, context)
 
+def _get_token():
```
ajinabraham: Nice !

StaticAnalyzer/views/windows.py
```diff
@@ -199,53 +222,98 @@ def _binary_analysis(tools_dir, app_dir):
...
-        bin_an_dic = __binskim(name, bin_an_dic)
-        bin_an_dic = __binscope(name, bin_an_dic)
+    if settings.CURRENT_PLATFROM != 'Windows':
+        if settings.WINDOWS_VM_IP is not None:
+            print "[INFO] Windows VM configured."
+            name = _upload_sample(bin_path)
+            bin_an_dic = __binskim(name, bin_an_dic)
+            bin_an_dic = __binscope(name, bin_an_dic)
+        else:
+            print "[INFO] Windowsc VM not configured in settings.py. Skipping Binskim and Binscope."
```
ajinabraham: Typo

StaticAnalyzer/views/windows.py
```diff
@@ -199,53 +222,98 @@ def _binary_analysis(tools_dir, app_dir):
...
-            "status": "Info",
-            "desc": "VM is not configured. Please read the readme.md in MobSF/install/windows."
-        }
-        bin_an_dic['results'].append(warning)
+        # TODO(Anpassen für windows lokal)
+        print "[INFO] Running lokal analysis."
+
+        global config
+        config = configparser.ConfigParser()
+        config.read('C:\\MobSF\\Config\\config.txt')
```
ajinabraham: Can you configure the config file location in settings.py, like https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/blob/master/MobSF/settings.py#L50

What can happen is that users may not always place MobSF under C:\ and also might not have permissions to create file there.

install/windows/config.txt
```diff
@@ -3,6 +3,8 @@ dir = C:\MobSF
 subdir_downloads = C:\MobSF\Downloads\
```
ajinabraham: Is this config applicable only for the Windows VM or for the Host Windows Machine as well?
If so we need to make if similar to 
https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/blob/master/MobSF/settings.py#L50

install/windows/readme.md
```diff
@@ -3,16 +3,17 @@
...
 
 ##Caution
-Use and separate Windows-VM for MobSF and *don't* expose it to a network range, where an attack might be coming from. The best solution is to set it to host-only mode.
+Use and separate Windows-VM for MobSF and *don't* expose it to a network range where an attack might be coming from. The best solution is to set it to host-only mode.
 
 ##Steps on the Windows-VM
 1. Install the following requirements on the VM
   * [Python 3](https://www.python.org/downloads/)
-  * Flask (via `python -m pip install flask`)
+  * Flask and rsa (via `python -m pip install flask rsa`)
```
ajinabraham: if the analysis is done on Host Windows. flask and rsa needs to go into `requirements.txt`

ajinabraham:
Code Review Done, Please look into the minor comments above.
Overall Looks Great 👍 

Will test the code next week and get back to you.

StaticAnalyzer/tools/strings.py
DominikSchlecht: Hmm I could specify `min` in `settings.py`.

Agreed on the special char, will implement this.

ajinabraham: We don't really want strings like

```
 B*o& 
IN\D[
yT((w
)]Xbj=
	l8k}
p'
  W

^
9SZj
?s9C
"P`(>
5t3|
]x4
30KwK
```

StaticAnalyzer/views/shared_func.py
DominikSchlecht: Ah ok, didn't see that.

Will be implemented.

StaticAnalyzer/views/windows.py
DominikSchlecht: Of course, will be implemented.

StaticAnalyzer/views/windows.py
DominikSchlecht: Yep, but that's a general problem.

Since I wanted to reuse setup.py from the windows-vm `C:\MobSF` with the corresponding configfile, settings.py is not used.

I could also specify the options in MobSFs settings.py, but if something changes (like a download link etc.) it would have to be changed in 2 locations. Like this, all windows-related config options are in /install/windows/config.txt and it costs minimum effort to keep windows-analysis running in the vm and locally.

If it's only the problem with the permissions under `c:\`, I could move the config to an user-folder. Would that satisfy the requirements or should I work out a more advanced solution with settings.py?

I could also load settings.py on the windows-analysis-vm and use it instead of `config.txt`, but I don't know if the mash-up with is really better than the current situation. A lot of unused options will be configured for local and remote analysis.

install/windows/config.txt
DominikSchlecht: Both :), see comment above.

StaticAnalyzer/views/windows.py
ajinabraham: > Since I wanted to reuse setup.py from the windows-vm C:\MobSF with the corresponding configfile, settings.py is not used.

I agree that you shouldn't be using settings.py inside Windows VM. 
But for the Host Windows Configuration It should as simple to the user that, they can make it work just by doing some config in `settings.py`.(If that is possible)

It's ok to place the config anywhere in the Windows VM.
But when it comes to MobSF running to Host OS. The config should be placed under
MobSF/MobSF/config.txt
or under `<User Home Dir>/.MobSF/config.txt` based on this: 
https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/blob/master/MobSF/settings.py#L33

https://github.com/ajinabraham/Mobile-Security-Framework-MobSF/blob/master/MobSF/settings.py#L65-L71

Or merged with settings.py (Highly recommend as this will be easier for User)
Let me know if I am missing something. 

FYI, `USE_HOME` setting is to support people to ship MobSF with custom OS/ Distros like Kali, Android Tamer etc. This settings allows to keep the config, db, uploaded files etc between MobSF dir or under user Home dir.



> I could also specify the options in MobSFs settings.py, but if something changes (like a download link etc.) it would have to be changed in 2 locations. Like this, all windows-related config options are in /install/windows/config.txt and it costs minimum effort to keep windows-analysis running in the vm and locally.


but at a time the user will either on VM or host Windows right?

DominikSchlecht: > Or merged with settings.py (Highly recommend as this will be easier for User)
Let me know if I am missing something. 

Sadly I really see a drawback - `setup.py` modifies `config.txt` to save paths of local progs like binskim (so we don't have to modify the path every time binskim jumps to a new version). As far as I understand, changes to `settings.py` made from MobSF are not persistent. (Unless I modify settings.py via loading and adding lines, but I don't think that this will be much better ^^)

> It's ok to place the config anywhere in the Windows VM.
But when it comes to MobSF running to Host OS. The config should be placed under
MobSF/MobSF/config.txt

So for now, I would implement this, since there are less changes required to the existing code.

> but at a time the user will either on VM or host Windows right?

yep, I meant the effort to keep 2 configs up-to-date for us (maintainers). A user will of course always use either the local one or the remote one.

ajinabraham: Okey then let it be like the existing implementation. I will test this setup and see if we can improve this.

install/windows/readme.md
```diff
@@ -8,7 +8,7 @@ Use and separate Windows-VM for MobSF and *don't* expose it to a network range w
 ##Steps on the Windows-VM
 1. Install the following requirements on the VM
   * [Python 3](https://www.python.org/downloads/)
-  * Flask and rsa (via `python -m pip install flask rsa`)
+  * rsa (via `python -m pip install flask rsa`)
```
ajinabraham: You might want to remove `flask` from command as well.

StaticAnalyzer/tools/strings.py
DominikSchlecht: Hmm, wrapping each result in ticks shows why these are shown:
```
'F
  Fe'
'2AF
'
'TopBorder'
'FarmKingHardCurrencyPackageX5'
'FRh0Fk'
'yVz`zjztz~z'
'z
{x{'
```

Newline seems to trick us here.
> string.printable
    String of characters which are considered printable. This is a combination of digits, letters, punctuation, and whitespace.

So:
```'0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~ \t\n\r\x0b\x0c'```
is the original, I would use:
```'0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~ '```

Result:
```
'0(0bpvp'
'FC`+h'
'9F*bhb"F(F'
'DiamondDiggerCollaborationSlots1Id5_Mobile'
'b0F!F'
'fakepurchase%d%d'
'v wt8'
'CandyMobileTencentColorBombIntroOffer'
'uAcid'
'FarmKingKakaoCampaignSoftCurrencyP1'
'wHxbx~x'
'FBF F$h9F'
'Self'
'tDt\txt'
```

File: install/windows/readme.md (hunk 9 above)
Comment (DominikSchlecht) [inline install/windows/readme.md]: Thanks, done with 2774e36

File: StaticAnalyzer/tools/strings.py (hunk 1 above)
Comment (ajinabraham) [inline StaticAnalyzer/tools/strings.py]: much better

File: StaticAnalyzer/tools/strings.py [hunk 10]
```diff
@@ -2,16 +2,15 @@
...
         result = ""
         for c in f.read():
-            if c in string.printable:
+            if c in '0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ!"#$%&\'()*+,-./:;<=>?@[\\]^_`{|}~ ':
                 result += c
                 continue
-            if len(result) >= min:
-                yield result
+            if len(result) >= min and result[0].isalnum():
+                yield "'" + result + "'"
```
ajinabraham: Do we need Single quotes?

DominikSchlecht: Thought they would be nice to spot whitespaces.

ajinabraham: fair enough

ajinabraham:
LGTM 👍 
Merging to `windows_static` branch so that I can test locally.

DominikSchlecht:
Great, thank you.

There are 2 paths hardcoded to my git repo:

- https://github.com/DominikSchlecht/Mobile-Security-Framework-MobSF/blob/master/install/windows/config.txt#L20

- https://github.com/DominikSchlecht/Mobile-Security-Framework-MobSF/blob/master/install/windows/setup.py#L20-L21

For testing these should be changed to your testing repo (e.g. https://raw.githubusercontent.com/ajinabraham/Mobile-Security-Framework-MobSF/windows_static/install/windows/config.txt).

A fix saving the files in the MobSF or user-path when installing on windows is already in my repository.

ajinabraham:
Nice! @DominikSchlecht updated the paths. Can you send the PR for user path support?

PR #237, opened 2016-09-30
[Windows-Static-Analysis] Storage of config and Tools in MobSF/User-Path

DominikSchlecht:
Rewrote some files so the tools and config/lock-files are stored in the MobSF-Path or the User-Path.


ajinabraham:
LGTM 👍 Merging

PR #320, opened 2017-01-17
Create binary_analysis.py

shuxin:
add elf check.


shuxin:
add elf check but without template.

PR #327, opened 2017-01-20
Estaban's Contribution Part 2

ajinabraham:
APKiD and Enhanced Permission Analysis


ajinabraham:
APKiD Added Separately

PR #348, opened 2017-02-19
iOS insecure connections

DominikSchlecht:
Hey Ajin,

I implemented a detection of custom [connection-settings](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html#//apple_ref/doc/uid/TP40009251-SW33) in Info.plist. Since the newer iOS-Versions won't allow insecure (HTTP or <TLS1.2) connections to happen if they are not specified in the Info.plist, I think this might add value to the framework.

- Implemented detection in iOS-View
- Updated HTML-Frontend (please check if the AF-Icons are the correct ones :))
- Updated PDF-Export

Could you please review and test the changes? Did already test with an IPA-File with multiple, single and none custom entries. Thank you in advance.

Greetings,
Dominik


ajinabraham:
Nice Addition, I will take it from here.

ajinabraham:
@DominikSchlecht 
I have added some more info, so that users will understand it better, remove `{ | safe}` may cause XSS, added icon, bumped a version of django. And tested it. 

Please merge if everything looks good. I am writing a minimal unit test suite so that it will be easy for everyone to test their integrations.

DominikSchlecht:
Looks good to me 👍

PR #377, opened 2017-03-23
Dockerfile updated. Now with sqlite client and pdf generator

OscarAkaElvis:
Now, the Dockerfile in the project acts in the same way the docker container you already have in dockerhub. Now it has sqlite client and pdf generator.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:16.04
 MAINTAINER Ajin Abraham <ajin25@gmail.com>
 
 ENV PDFGEN_PKGFILE wkhtmltox-0.12.1_linux-trusty-amd64.deb
-ENV PDFGEN_URL http://download.gna.org/wkhtmltopdf/0.12/0.12.1/${PDFGEN_PKGFILE}
+ENV PDFGEN_URL https://downloads.wkhtmltopdf.org/0.12/0.12.1/${PDFGEN_PKGFILE}
 
 #Update the repository sources list
 RUN apt-get update -y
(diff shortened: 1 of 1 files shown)


ajinabraham:
Can you replace 'http://download.gna.org/wkhtmltopdf/0.12/0.12.1/' with 'https://bitbucket.org/wkhtmltopdf/wkhtmltopdf/downloads/wkhtmltox-0.13.0-alpha-7b36694_linux-trusty-amd64.deb'

Good to have `https`

OscarAkaElvis:
Ok, but changing the file from `wkhtmltox-0.12.1_linux-trusty-amd64.deb` to `wkhtmltox-0.13.0-alpha-7b36694_linux-trusty-amd64.deb`, the dependencies are different. Now with that file you propose, it needs a package called libicu52 which doesn't exist in Ubuntu Xenial (remember we are using FROM ubuntu:16.04 on docker file). In Xenial exists libicu55, I tested installing it but the pdf generator is still asking for libicu55, so this needs to be checked deeply to use the file you propose.

ajinabraham:
My bad I recommended arm before. It should be amd :D
@OscarAkaElvis You can use `wkhtmltox-0.12.1_linux-trusty-amd64.deb` in that case. My only concern is that the URL should be `https`

Can you try this one instead: `https://ftp.de.debian.org/debian/pool/main/w/wkhtmltopdf/wkhtmltopdf_0.12.3.2-3_amd64.deb`

OscarAkaElvis:
With that file there are dependency changes too. If you are concerned only with https, I found a url (in fact is the official) with https and the file I'm using in Dockerfile. Is: https://downloads.wkhtmltopdf.org/0.12/0.12.1/wkhtmltox-0.12.1_linux-trusty-amd64.deb

So I'll push in minutes the change after testing if everything is working.

OscarAkaElvis:
Checked! and it's working fine. Push done.

ajinabraham:
LGTM 👍

PR #416, opened 2017-05-19
solve error in docker container

superpoussin22:
same erros as kali so need to apply kali patch when building image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,8 +2,8 @@ FROM ubuntu:16.04
 
 LABEL maintainer="Ajin Abraham <ajin25@gmail.com>"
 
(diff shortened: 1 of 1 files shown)


ajinabraham:
@superpoussin22 Kali Fix was to enable 32 bit execution support. What component need 32 bit execution support for Static Analysis ?

superpoussin22:
just saw this warning at container startup "[WARNING] You don't have 32 bit execution support enabled or MobSF shipped ADB binary is not compatible with your OS.
Please set the 'ADB_BINARY' path in settings.py"

so i try to solve it, but perhaps not needed :)

ajinabraham:
Yes we don't need it, but I will merge the PR so that the error won't show up.

Issue #415, opened 2017-05-18
Static Analysis - Malware Check - False Positive


SatanicDrNull:
## EXPLANATION OF THE ISSUE

On version 0.92 as well 0.95 I detected the framework incorrectly matches the IP 127.0.0.1 with a string inside the Malware-DB
This was verified on android component. Inspecting the code, I am almost sure the same occurs on iOS and Windows

The problem seems to be on line 132 of the views.py file inside the MalwareAnalyzer folder:

`if (domain in details_dict["domain_or_url"]) or (domain in details_dict["ip"]):`

This part of the code tells to seek the  "domain" content  inside the "details_dict['ip']

As the string "127.0.0.1" is part of the registry on MalwareDb file: 
`193.107.19.80/inter/gate.php?hwid=4281525696&pc=COMPUTERNAME&localip=127.0.0.1&winver=x32`

It returns as a false positive.
```

 
## STEPS TO REPRODUCE THE ISSUE

```
1. Upload an Apk where inside the code is referred the url "http:\\127.0.0.1"
2. Execute the static analysis
3. The Malware Check section will show a malware is found.
```

## CONSOLE OUTPUT

![bug_mobsf](https://cloud.githubusercontent.com/assets/28790490/26220208/1dc03c60-3be8-11e7-84d3-9e04b9b0da77.JPG)

## CONTENTS OF LOG FILES

```
Paste the contents of logs/MobSF.log here
```
No MobSF.log was generated.
```
If you have issues with API Fuzzer,
Paste the contents of logs/webproxy.log here
(NOT REQUIRED FOR FEATURE REQUEST/ENHANCEMENT)
```
<!--
BEFORE POSTING YOUR ISSUE/BUG
- These comments won't show up when you submit the issue.
- GitHub issues ARE NOT FOR SUPPORT, DISCUSSIONS AND QUESTIONS! 
- If you have questions, use our slack channel. Request invitation here https://goo.gl/6cYU5a.
- Installation problems with regards to requirements.txt or VirtualBox in not considered as a MobSF issue.
- Reproduce issue in the latest master and try to add as much detail as possible.
- If you're requesting a new feature/enhancement, explain why you'd like it to be added and it's importance.
- Search this repository (top of the page) for the issue and it has not been fixed or reported already.
- Once you open a bug, you should also provide additional information if requested. 
- Failure to do so in 25 days will result in closure of the bug without further communication.
-->

[...]

PR #431, opened 2017-06-20
correction1

PaulAnton:
taking the part of the IP before the forward slash- this way it will avoid the part that might contain 127.0.0.1. what would be the format of the urllist? assuming it will have just the part before the forward slash- this code should work fine


PaulAnton:
#415

ajinabraham:
Reviewing and Testing Code.

ajinabraham:
I did an alternate fix.

ajinabraham:
Fixed in https://github.com/MobSF/Mobile-Security-Framework-MobSF/commit/9aed2b8b3328d7ae001fce90a87242cc71eae0d1

PR #438, opened 2017-07-11
Saving the dynamic results

matandobr:
Once done, the dynamic results should be accessible again

### DIFF
--- a/StaticAnalyzer/views/android/static_analyzer.py
+++ b/StaticAnalyzer/views/android/static_analyzer.py
@@ -208,6 +208,7 @@ def static_analyzer(request):
                         bin_an_buff,
                         apkid_results,
                     )
+                context['dynamic_analysis_done'] = os.path.exists(os.path.join(app_dic['app_dir'], 'logcat.txt'))
                 template = "static_analysis/static_analysis.html"
                 return render(request, template, context)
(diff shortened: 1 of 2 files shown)


ajinabraham:
Is this under dev?

matandobr:
I think it's working fine, except there is no way to delete the results from the ui now, I can add this functionality inside the static "reanalyze" function, or to separate it. What do you think?

ajinabraham:
@matandobr Cool. I will do a review of the code. 
I think how it works now is by overwriting the previous scan results which is fine for dynamic analyser. The delete function needs to be implemented to the page that shows the scan result.

matandobr:
Okay not a problem, I'll send an update

ajinabraham:
Also where we list the Recent Scan, IIRC.

ajinabraham:
Next time you make a PR. It will be easy if you create a branch from the main repo.

ajinabraham:
@matandobr I think is good and easy hack to show the dynamic analysis report.
Don't add to Recent Scan. But a delete function that will extend the delete button in `http://127.0.0.1:8000/recent_scans/` would be great. essentially it should delete the db scan results for static analysis and the files created during dynamic analysis.

Issue #412, opened 2017-05-18
Android x86 Found elf built without Stack Protection Question


maxwell-nc:
Accroding the Hint Built with option -fstack-protector, but no effect with x86 platform.
so i find the source code ,which i get:
```python
            if section_header['sh_entsize'] > 0:
                for i in range(section_header['sh_size'] // section_header['sh_entsize']):
                    entry = elffile.decode_sym(
                        section_header['sh_offset'] + i * section_header['sh_entsize'])
                    name = elffile.decode_string(
                        strtab_section + entry['st_name'])
                    if name == '__stack_chk_guard':
                        has_sp = True
```
but i when i print “name ” on x86 platform, 
i find "__stack_chk_fail" like "__stack_chk_guard",

if i use flag += -fno-stack-protector,
i also find "__stack_chk_fail" in x86's "name".

Is there something wrong with the judgement, or just my mistake?
how can i fix this problem? thanks.

ajinabraham:
This code was contributed by @shuxin.
@shuxin Can you take a look on this one?

doomguy:
@maxwell-nc
Seems like sometimes 3rd party libraries pull Stack Protection into the binary. You can try to guess where it came from using something like this:

`objdump -d mylibrary.so | egrep '>:$|__stack_chk_fail' | grep -B1 __stack_chk_fail`

Also have a look at https://stackoverflow.com/questions/30532327/gcc-fno-stack-protector-for-linking-not-work for some more info.

@shuxin 
I'm still unsure if it is wise to rely solely on the presence of __stack_chk_guard to determine whether Stack Protection is enabled or not. Tools like checksec.sh (http://www.trapkit.de/tools/checksec.html) just look for __stack_chk_fail. Other ressources also seem to concentrate on __stack_chk_fail for detection (https://www.win.tue.nl/~aeb/linux/hh/protection.html). Can you maybe clarify why you choose __stack_chk_guard in the first place?

Thanks.

ajinabraham:
Related Issue: https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/454

shuxin:
@doomguy 
Thanks.
I used __stack_chk_guard instead of __stack_chk_fail because crtbegin_dynamic.o was linked when the abi was x86. I apologize for the imperfect implementation. I'll fix some.
But you know, there is not a good way to check every linked part in an whole elf file. I'm looking forward to a better solution.

[...]

PR #470, opened 2017-08-24
fix for Issue #412

shuxin:


ajinabraham:
Thanks for the patch @shuxin

PR #485, opened 2017-09-23
VirusTotal scan report

matandobr:
Optional feature using the public api (free)
see settings.py for additional info


ajinabraham:
Code Looks good to me. Please see my comments. 
Testing it out.

ajinabraham:
Also in the result, if the file is not infected, will VirusTotal send a list of AV and corresponding status. 
It would be nice to display these AVs and a clean span element in that case.

matandobr:
> Also in the result, if the file is not infected, will VirusTotal send a list of AV and corresponding status.
It would be nice to display these AVs and a clean span element in that case.
 In case of infection detected, does the result contain all AVs that didn't detected this as a threat?

No matter what the results are, VT returns a list of all the results from all the AVs.
The thing is they are not saying anything useful when they don't find anything malicious,  So, I could display all the "clean" results but I figured I would prefer to filter those out... 
I filter the clean results in the template itself :
`{% if av_result.detected == True %}`

ajinabraham:
My only concern is that when there is a false positive like this
screen shot 2017-09-24 at 12 10 09 pm
It might give a false sense to some users. If we can also show that other reputable AVs show clean. Then the user can go ahead with bit more confidence on the file. There are lot of generic things in a binary that can trigger a possible signature even when the file is perfectly safe. (Crappy AV logic)

ajinabraham:
Also when there is a response like this in the console.
```
[INFO] VirusTotal: Check for existing report
[INFO] VirusTotal: The requested resource is not among the finished, queued or pending scans
[INFO] VirusTotal: file upload
[INFO] VirusTotal: Scan request successfully queued, come back later for the report
```

Can you share `Scan request successfully queued, come back later for the report` in the Web GUI as well?

matandobr:
> My only concern is that when there is a false positive like this
You're right, on the other hand, this is no better:
image
The problem here it's creating a HUGE table with no relevant information, and from my experience, it's pretty common to filter these results out.
So I think this way is better because it's cleaner 
What do you think?

ajinabraham:
May be a balance is better, like when the detection is less than 5, you should show the rest of the av results and you can show something like Clean tag instead of None. There is lot of crappy AVs in VT that get triggered very easily. We should not give a false indicator to user even when 44 other AV says it's clean.

matandobr:
> May be a balance is better, like when the detection is less than 5, you should show the rest of the av results and you can show something like Clean tag instead of None. There is lot of crappy AVs in VT that get triggered very easily. We should not give a false indicator to user even when 44 other AV says it's clean.

On the other hand, sometimes a just-published-malware will not be signed in all the engines so we could be wrong either way, I think we should let the user (which I hope is having some kind of knowledge in the security world) make his on choices...

ajinabraham:
Well then, let's push it that way.

matandobr:
(Where is the open-source community when it needed?!) ;)

matandobr:
> Can you add Scan request successfully queued, come back later for the report under Virus Total Report in the Web GUI as well?

Working on it

ajinabraham:
Can you put detection inside `<span class="label label-danger">desc</span>` and a link to VirusTotal for users who want to see more details. So that they can get the full results if they want. (Does VT allows direct linking).

matandobr:
@ajinabraham waiting for your comments in the slack chat

ajinabraham:
@matandobr Did you miss adding VT results to PDF report for Android binaries? I don't see a android static analysis PDF template in the commit list.

PR #515, opened 2017-12-05
Fixed #476

### LINKED ISSUE #476: [ Possible BUG or enhancement needed in MOBSF] Certificate and Signing
Hi,

Just a little question regarding MobSF and apk signature.

you can use an SHA1 certificate to sign the apk but use sha256 for signing the apk .

Mobsf will see that the certificate use SHA1 ans report a bad state which is good regarding the certificate  but don't reflect how files are hashed.

I think mobsf should inspect the MANIFEST.MF file or better ht .SF to have apksigner information

What do you think ?

PS: What will define the signature algorithm used by apksigner is mostly the MinSDK version

ramimac:
Passes all tests locally and correctly provides guidance on potential false positives


ajinabraham:
Testing this out.

ajinabraham:
Hi @ramimac Can you remove
StaticAnalyzer/migrations/0002_auto_20171128_1451.py
StaticAnalyzer/migrations/0002_auto_20171128_1451.py/
from the commit.

PR #542, opened 2018-02-05
Update to Apktool 2.3.1 and variable

superpoussin22:
make apktool version a variable readed from settings.py so no more need to chnage the code in manifest_analysis.py

only need to upload the new apktool version and change value in settings.py


superpoussin22:
my bad it contain commit for upstream proxy :(

changes files are only settings.py and manifest_analysis.py

will now create a branche for each chnage and no more work on master 

sorry

PR #543, opened 2018-02-05
Python3 apktool

superpoussin22:
make apktool version a variable in settings.py


superpoussin22:
no need so I close it

PR #547, opened 2018-02-08
upgrade maximum upload file size

superpoussin22:
Maximum upload fiel size is set to 100MB

recent ipa are  bigger than this perhaps we could increase to 300 ?

In particular because that's not easy for regular user to find where this value is stored


ajinabraham:
I agree, would be nice to read it from Settings.py

ajinabraham:
LGTM 👍 Good to merge

superpoussin22:
will work on adding it in settings.py

PR #548, opened 2018-02-08
update about page

superpoussin22:
add more info about MobSF and what it can do


ajinabraham:
Add a small commit to include other core devs.
👍  LGTM good to merge

PR #549, opened 2018-02-09
Replace strings_from_apk.jar with androguard string extraction

superpoussin22:
Tested on MacOS, Ubuntu, Windows 7 and with Docker image

as requested androguard is installed with pip and not with a git clone 
(ubuntu do have a package)

@ajinabraham need some test 

Comments are welcome

### DIFF
--- a/StaticAnalyzer/views/android/strings.py
+++ b/StaticAnalyzer/views/android/strings.py
@@ -2,8 +2,8 @@
 """Module for strings-method for java."""
 
 import io
-import os
 import subprocess
+from androguard.core.bytecodes import apk
 
 from django.conf import settings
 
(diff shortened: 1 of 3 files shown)


ajinabraham:
@superpoussin22 Should be nice if you give a meaning ful title and description in the PR like. This is useful for project management. See: https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/python3/.github/PULL_REQUEST_TEMPLATE.md

requirements.txt
```diff
@@ -6,3 +6,5 @@ configparser==3.5.0
 pdfkit==0.6.1
 lxml==4.1.1
 requests==2.18.4
+ipython>=5.0.0,<6
```
ajinabraham: Is this required?

ajinabraham:
@superpoussin22 Can you add a commit to remove `strings_from_apk.jar` in this PR.

StaticAnalyzer/views/android/strings.py
```diff
@@ -12,30 +12,25 @@
...
-                encoding="utf8",
-                errors="ignore"
-            ) as file_pointer:
-                dat = file_pointer.read()
-        except:
-            pass
-        dat = dat[1:-1].split(",")
+        print "[INFO] Extracting Strings from APK"
+
+        a = apk.APK(app_dir+app_file)
```
ajinabraham: Let's use `os.path.join()`

requirements.txt
superpoussin22: ipython is used by androguard, it should be installed by androguard , but I got a case where it wasn't so I add it by security

superpoussin22:
@ajinabraham done for remove strings_from_apk.jar

requirements.txt
ajinabraham: Can you please confirm if it's not installed by androguard and is a requirement for androguard strings extraction function to work?

superpoussin22: Normally we could remove it from requierements.txt   will redo all the tests to confirm , it is correctly installed by androguard

superpoussin22: we can remove ipython>=5.0.0,<6 in the requirement list ak-androguard do install it correctly 

will update PR

StaticAnalyzer/views/android/strings.py
superpoussin22: will do it :)

superpoussin22: will do

requirements.txt
superpoussin22: will update the file , ak-androguard do install it correctly no need to add it, iw will introduce a risk if we forget to update it if needed

StaticAnalyzer/views/android/strings.py
```diff
@@ -12,30 +13,27 @@
...
-                app_dir + 'strings.json',
-                mode='r',
-                encoding="utf8",
-                errors="ignore"
-            ) as file_pointer:
-                dat = file_pointer.read()
-        except:
-            pass
-        dat = dat[1:-1].split(",")
+        print "[INFO] Extracting Strings from APK"
```
ajinabraham: Problem!

StaticAnalyzer/views/android/strings.py
```diff
@@ -12,30 +13,27 @@
 )
 
 
-def strings_jar(app_file, app_dir, tools_dir):
+def strings(app_file, app_dir, tools_dir):
```
ajinabraham: Rename this to `strings_jar`
else this will break the code
https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/python3/StaticAnalyzer/views/android/static_analyzer.py#L43

ajinabraham: remove `tools_dir` from here as well as from the caller. It's not required.

ajinabraham:
I made the changes

PR #550, opened 2018-02-12
Correct Android manisfet xml output

superpoussin22:
Since we go to Python 3 , the output of the android manifest was not correct due to the byte string output of python3 ( b'<xml  instead of <xml)

 #dat = dat.replace(b"\n", b"")
 dat = dat.decode()

### DIFF
--- a/StaticAnalyzer/views/android/manifest_analysis.py
+++ b/StaticAnalyzer/views/android/manifest_analysis.py
@@ -1269,7 +1269,8 @@ def read_manifest(app_dir, tools_dir, typ, binary):
 
             args = [settings.JAVA_PATH + 'java', '-jar', cp_path, manifest]
             dat = subprocess.check_output(args)
-            dat = dat.replace(b"\n", b"")
+            #dat = dat.replace(b"\n", b"")
+            dat = dat.decode()
         else:
             print("[INFO] Getting Manifest from Source")
             if typ == "eclipse":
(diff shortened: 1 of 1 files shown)


StaticAnalyzer/views/android/manifest_analysis.py
```diff
@@ -1269,7 +1269,8 @@ def read_manifest(app_dir, tools_dir, typ, binary):
 
             args = [settings.JAVA_PATH + 'java', '-jar', cp_path, manifest]
             dat = subprocess.check_output(args)
-            dat = dat.replace(b"\n", b"")
+            #dat = dat.replace(b"\n", b"")
+            dat = dat.decode()
```
ajinabraham: What about? 
`dat.decode("utf-8", "ignore") `

superpoussin22: if you re-add dat.replace(b"\n", b"")
it will produce
image

ajinabraham: I understand, I am asking about line 1273. Shall we use `dat.decode("utf-8", "ignore")` ? Because that looks more failsafe. what do you think?

superpoussin22: agree with you to  use dat.decode("utf-8", "ignore")

PR #555, opened 2018-02-23
Dockerfile with APKiD enabled

ajinabraham:
Enable and configure APKiD in Dockerfile
By @superpoussin22


ajinabraham:
@superpoussin22 Looking Good. Merging this!

PR #563, opened 2018-03-13
Fix manifest reading

superpoussin22:
read AndroidManifest.xml for apktool if file exist

### DIFF
--- a/StaticAnalyzer/views/android/manifest_analysis.py
+++ b/StaticAnalyzer/views/android/manifest_analysis.py
@@ -1253,16 +1253,15 @@ def manifest_analysis(mfxml, man_data_dic):
         PrintException("[ERROR] Performing Manifest Analysis")
 
 
-def read_manifest(app_dir, tools_dir, typ, binary):
+def read_manifest(app_dir, tools_dir, typ, apk):
     """Read the manifest file."""
     try:
         dat = ''
         manifest = ''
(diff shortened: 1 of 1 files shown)


StaticAnalyzer/views/android/manifest_analysis.py
```diff
@@ -1260,8 +1260,21 @@ def read_manifest(app_dir, tools_dir, typ, binary):
...
+                    encoding="utf8",
+                    errors="ignore"
+                ) as file_pointer:
+                    dat = file_pointer.read()
+                return dat
+            else:
+                print("[INFO] Using Manifest from Binary")
+                print("[INFO] AXML -> XML")
+                manifest = os.path.join(app_dir, "AndroidManifest.xml")
             if len(settings.AXMLPRINTER_BINARY) > 0 and isFileExists(settings.AXMLPRINTER_BINARY):
```
ajinabraham: What is the scope of this else ? In this case both code using apktool and axmlprinter executes as if at line 1278 is in outer scope. We don't need both to execute, right?

superpoussin22: you have a return on line 1273 if apktool_out/ contain a Manifest 

will review the code, Cause I thins the if in 1278 should be "idented" better  part of the else on 1274 

good catch @ajinabraham

PR #590, opened 2018-04-10
Update icon_analysis.py

superpoussin22:
1st correction . need now to understand why the Hidden icon image is not displayed

### DIFF
--- a/StaticAnalyzer/views/android/icon_analysis.py
+++ b/StaticAnalyzer/views/android/icon_analysis.py
@@ -110,8 +110,9 @@ def get_icon(apk_path, res_dir):
     try:
         print("[INFO] Fetching icon path")
         a = apk.APK(apk_path)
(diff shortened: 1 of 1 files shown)


ajinabraham:
@matandobr Do add a review comment.

superpoussin22:
@matandobr  xml file is an adaptive icon, androguard team is working on it, so, my test sounds Ok until they found a solution 

https://github.com/androguard/androguard/pull/454

matandobr:
So we leave it like that for now? We don't have the hidden icon anomaly anymore and that sucks..

superpoussin22:
@matandobr  perhaps we could open a ticket for hidden icon anomaly and if you have some apk hash I can use to diagnose the issue, I will work on it  but merge this PR to enhance icon display ?

ajinabraham:
LGTM 👍

PR #618, opened 2018-05-01
AVD android 6 update

matandobr:
Finally :)

* Major update to the AVD emulator 
* few aesthetic changes

- Don't merge this PR untill the image zip is uploaded !


MobSF/settings.py
```diff
@@ -345,12 +347,9 @@
...
 
     #===========ANDROID EMULATOR SETTINGS ===========
-    # generated by mobsfy_AVD.py, do not edit the
-    # below AVD settings yourself.
-    AVD_EMULATOR = "avd_emulator"
-    AVD_PATH = "avd_path"
-    AVD_REFERENCE_NAME = r'Nexus5API16'
-    AVD_DUP_NAME = r'Nexus5API16_1'
+    # Android-Studio 'emulator' binary path
+    AVD_EMULATOR = "/Users/matandobr/Library/Android/sdk/tools/emulator"
```
ajinabraham: Do you want to provide this as a commented example and set the default to `""`

MobSF/utils.py
```diff
@@ -468,6 +465,9 @@ def gen_sha256_hash(msg):
 def isFileExists(file_path):
     if os.path.isfile(file_path):
         return True
+    # This fix situation where a user just typed "adb" or another executable inside settings.py
+    if shutil.which(file_path):
```
ajinabraham: works on windows?

scripts/start_avd.py
```diff
@@ -0,0 +1,102 @@
+import sys
+import os
+import platform
+import subprocess
+import django
+from os.path import dirname, abspath
+
+MobSF_path = dirname(dirname(abspath(__file__)))
```
ajinabraham: Is this already available from `settings.py`?

scripts/start_avd.py
```diff
@@ -0,0 +1,102 @@
...
+sys.path.append(MobSF_path)
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'MobSF.settings')
+django.setup()
```
ajinabraham: Why these code?
From line 8-11 ?

scripts/start_avd.py
```diff
@@ -0,0 +1,102 @@
...
+        return True
+    except:
+        PrintException("[ERROR] help_boot_avd")
+        return False
+
+
+def start_avd():
+    try:
+        if platform.system() == 'Darwin':
+            # There is a strage error in mac with the dyld one in a while..
```
ajinabraham: once I guess.

MobSF/settings.py
matandobr: Mmm doesn't matter, whatever you think is better (I check if the binary exists in the start_avd)

MobSF/utils.py
matandobr: Didn't test on Windows & Linux yet
(I guess I will send some fixes after I'll do it)

scripts/start_avd.py
matandobr: What do you mean?

scripts/start_avd.py
matandobr: So I could use the Django settings interface (instead of parsing settings.py manually or passing arguments into the script)
It basically "starting" the Django engine of the project, just so I could use the setting

scripts/start_avd.py
matandobr: Yup

MobSF/settings.py
```diff
@@ -292,7 +294,7 @@
     JD_CORE_DECOMPILER_BINARY = ""
     PROCYON_DECOMPILER_BINARY = ""
     APKTOOL_BINARY = ""
-    ADB_BINARY = ""
+    ADB_BINARY = "adb"
```
ajinabraham: How about defaulting to appropriate version from `/DynamicAnalyzer/tools/adb/` ?

scripts/start_avd.py
```diff
@@ -0,0 +1,102 @@
...
+def main():
+    print("[INFO] MobSF - start_avd.py has started")
+    if not check_config():
+        return -1
+    if not start_avd():
+        return -1
+    if not help_boot_avd():
+        return -1
+    print("[INFO] start_avd successfully finished")
+    print("[INFO] Please wait untill the emulator will load completely, only then take a snapshot")
```
ajinabraham: Typo `until`

zseha:
Any news on this?

zseha:
No need to be sorry. I was just wondering if you needed some help?

matandobr:
@zseha 
Well, help is always welcomed, 
The thing is it's more of a research task,
I'm trying delivering a whole avd folder, with modified system image inside of it so the deployment will be good too and I guess some of the configurations of the system image is static to the host which downloaded it or something like this.. 
Hope I'll have a breakthrough soon

rohandora:
@zseha thanks for pointing out,will wait till this gets merged.

PR #641, opened 2018-05-30
Docker postgres

superpoussin22:
Add a new Dockerfile for postgres support + an docker-compose file to be able to build Mobsf Docker image with postgres support and launch both services.


superpoussin22:
will work on this :)

superpoussin22:
@ajinabraham  let me know if better for you :)

Dockerfile
```diff
@@ -74,6 +74,9 @@ RUN ./kali_fix.sh
 WORKDIR /root/Mobile-Security-Framework-MobSF
 RUN pip3 install -r requirements.txt
 
+#install Postgres dependencies
+RUN pip3 install psycopg2
```
ajinabraham: Is this required here? 
Should this be moved to `Dockerfile.postgres`?

superpoussin22: yep agree

superpoussin22: removed

ajinabraham:
@superpoussin22 persistence was not mandatory but an optional feature like normal one with and without persistence.

Also I am a bit worried about `Dockerfile.postgres` Now we need to maintain both `Dockerfile` and `Dockerfile.postgres`

Can we use build time args to support postgres and non postges version, with/without persistence ?

I see
```
Set build-time variables (--build-arg)
You can use ENV instructions in a Dockerfile to define variable values. These values persist in the built image. However, often persistence is not what you want. Users want to specify variables differently depending on which host they build an image on.

A good example is http_proxy or source versions for pulling intermediate files. The ARG instruction lets Dockerfile authors define values that users can set at build-time using the --build-arg flag:

$ docker build --build-arg HTTP_PROXY=http://10.20.30.2:1234 .
This flag allows you to pass the build-time variables that are accessed like regular environment variables in the RUN instruction of the Dockerfile. Also, these values don’t persist in the intermediate or final images like ENV values do.

Using this flag will not alter the output you see when the ARG lines from the Dockerfile are echoed during the build process.

For detailed information on using ARG and ENV instructions, see the Dockerfile reference.
```
More info: https://docs.docker.com/engine/reference/commandline/build/ and navigate to `Set build-time variables (--build-arg)` section.

superpoussin22:
will take a look if we can use variable :)

superpoussin22:
it now use variable for postgres support and by default postgres support is set to False

README.md
```diff
@@ -39,6 +39,11 @@ docker run -it -p 8000:8000 opensecurity/mobile-security-framework-mobsf:latest
 ```
 docker run -it -p 8000:8000 -v <your_local_dir>:/root/.MobSF opensecurity/mobile-security-framework-mobsf:latest
 ```
+### For postgres support
```
ajinabraham: we will move this to docker wiki and add a link to it
`https://github.com/MobSF/Mobile-Security-Framework-MobSF/wiki/7.-Docker-Container-for-MobSF-Static-Analysis `

ajinabraham:
LGTM 👍

PR #680, opened 2018-08-09
New apkid

superpoussin22:
Yara apkid do have a new dex support and also a new rules format


superpoussin22:
test done with a few apk and apk including bad things in it :)

PR #682, opened 2018-08-14
iOS IPA Analysis support inside Docker and Linux

ajinabraham:
### What was a problem?

```
We were not able to perform iOS iPA analysis inside docker/Linux
due to lack of tooling support like otool
```

```
We have a work around with the help of jtool.
Thanks to http://www.newosxbook.com/tools/jtool.html
```

- [x] Run MobSF unit tests

```
Tested in Ubuntu, Mac and Docker
Won't work in Windows. Currently we support IPA analysis only in mac and linux.
```


moshwemoshwe:
Hi, What is the best way?

PR #685, opened 2018-08-15
api report json support 400 404 statuscode

magaofei:
### What was a problem?

```
add 400 and 404 status codes for Generate PDF Report API and Generate JSON Report API
```

```
==
when i input a error parameter for scan_type , like "fefr" , it return `[ERROR] local variable 'context' referenced before assignment`
```

- [x] Run MobSF unit tests

```
DESCRIBE HERE
```


ajinabraham:
@magaofei Could you please resolve conflicts?

magaofei:
@ajinabraham ok

magaofei:
And i want to rewrite some code of `MobSF.rest_api.py` , there are a lot of `IF_ELSE` statement, for example, you can use django form. Do you agree? I can try

ajinabraham:
Yes I agree. it needs code refactoring.
```
python -c "import this" | grep nest
Flat is better than nested.
```
👍

magaofei:
OK, i got it.

ajinabraham:
@magaofei Does it add too much value than the existing implementation? If not lets stick with the existing setup and avoid extra dependencies.

magaofei:
@ajinabraham ok

PR #695, opened 2018-08-22
Correct apkid result

superpoussin22:
### What was a problem?

```
some APKID results not reported DESCRIBE HERE
```

```
It analyze dex and also apk 
```

- [x] Run MobSF unit tests

```
It require optimization I think , in views.py
```

### DIFF
--- a/MalwareAnalyzer/views.py
+++ b/MalwareAnalyzer/views.py
@@ -33,7 +33,7 @@ def apkid_analysis(app_dir,app_file):
         else:  
             file_to_analyze = app_dir + 'classes.dex'
             print("[INFO] APKiD Analysis on Dex file")
-            if not os.path.exists(dex_file):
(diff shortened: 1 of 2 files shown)


MalwareAnalyzer/views.py
```diff
@@ -79,6 +79,39 @@ def apkid_analysis(app_dir):
         else:
             apkid_dict["manipulator"] = ""
 
+        apkid_dict["result"] = ""
```
ajinabraham: From here it's double work. @superpoussin22 It should like do scan of APK first, if it fails to do so try dex scan and not both.

superpoussin22: @ajinabraham for the moment scanning only the apk did not provide all the informations, need to check apid.py to undestand why , so far it's a quick fix

ajinabraham: okey then

PR #698, opened 2018-08-24
Request method decorator and update form errors

magaofei:
### What was a problem?

```
change the Django form error to
        {
        "error": {
            "file": "This field is required.", 
            "test": "This field is required."
            }
        }

the decorator of request method is easy to use
put the ['get'] or ['post'] on func name
```

```
DESCRIBE HERE
```

- [x] Run MobSF unit tests

```
DESCRIBE HERE
```

### DIFF
--- a/MobSF/forms.py
+++ b/MobSF/forms.py
@@ -26,7 +26,7 @@ class FormUtil(object):
         """
         data = form.errors.get_json_data()
         for k, v in data.items():
-            data[k] = '; '.join([value_detail['message'] for value_detail in v])
+            data[k] = ' '.join([value_detail['message'] for value_detail in v])
         return data
 
     @staticmethod
--- a/MobSF/forms.py
+++ b/MobSF/forms.py
@@ -16,14 +16,19 @@ class FormUtil(object):
         form.errors.get_json_data() django 2.0 or higher
 
         :return
-        example { "error": "file This field is required." }
+        example
+        {
+        "error": {
+            "file": "This field is required.", 
+            "test": "This field is required."
+            }
+        }
         """
-        errors_messages = []
-        for k, value in form.errors.get_json_data().items():
-            errors_messages.append(
-                (k + ' ' + ' , '.join([i['message'] for i in value])))
-        return '; '.join(errors_messages)
+        data = form.errors.get_json_data()
+        for k, v in data.items():
+            data[k] = '; '.join([value_detail['message'] for value_detail in v])
+        return data
 
     @staticmethod
     def errors(form):
-        return form.errors.items()
+        return form.errors.get_json_data()
--- a/MobSF/utils.py
+++ b/MobSF/utils.py
@@ -19,7 +19,8 @@ import shutil
 from . import settings
 
 from django.shortcuts import render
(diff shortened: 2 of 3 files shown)


MobSF/forms.py
```diff
@@ -16,14 +16,19 @@ def errors_message(form):
...
+        }
         """
-        errors_messages = []
-        for k, value in form.errors.get_json_data().items():
-            errors_messages.append(
-                (k + ' ' + ' , '.join([i['message'] for i in value])))
-        return '; '.join(errors_messages)
+        data = form.errors.get_json_data()
+        for k, v in data.items():
+            data[k] = '; '.join([value_detail['message'] for value_detail in v])
```
ajinabraham: `'; '` ?

ajinabraham: Will that affect this format?
```
   {
        "error": {
            "file": "This field is required.", 
            "test": "This field is required."
            }
        }
```

MobSF/utils.py
```diff
@@ -13,11 +13,16 @@
 import io
 import ast
 import unicodedata
+import functools
 import requests
 import shutil
 from . import settings
 
 from django.shortcuts import render
+from django.http import HttpResponseNotAllowed, HttpRequest
```
ajinabraham: We follow the styling as follows
```
from django.http import (
    HttpResponseNotAllowed,
   HttpRequest
)
```

MobSF/utils.py
```diff
@@ -630,4 +635,37 @@ def is_zip(self):
...
+        @functools.wraps(func)
+        def wrapper(*args, **kwargs):
+
+            if not isinstance(methods, list) and not isinstance(methods, tuple):
+                raise ValueError('the parameter methods is not a list or tuple')
+            
+            methods_upper = [m.upper() for m in methods]
+            for method in methods_upper:
+                if method not in ALLOW_METHODS:
+                    raise ValueError('the methods is not allow')
```
ajinabraham: Please change it to `This method is not allowed`

MobSF/utils.py
```diff
@@ -630,4 +635,37 @@ def is_zip(self):
...
+
+            request = None
+            for arg in args:
+                if isinstance(arg, HttpRequest):
+                    request = arg
+            if request is None:
+                raise ValueError('the parameter not have request')
```
ajinabraham: Change to `Request object not found`

MobSF/forms.py
magaofei: No, because the origin data is 
```json
{
  "error": {
    "file": [
      {
        "message": "This field is required.",
        "code": "required"
      }
    ],
    "test": [
      {
        "message": "This field is required.",
        "code": "required"
      }
    ]
  }
}
```

the value of file and test is list, that means there may be multiple values

MobSF/utils.py
magaofei: ok

MobSF/utils.py
magaofei: ok

MobSF/utils.py
magaofei: ok

PR #703, opened 2018-08-28
Update views.py

superpoussin22:
No need to launch apkid twice
also no need to specifiy the dex file cause we are launching apkid.scdan with parameters set to True so a directory is enough

### What was a problem?

```
previous version launch to apk id on one dex and one on apk to detect obfuscator....
```

```
Only one launch but thetre's to array not only one , so parsing also the second one 
```

- [X] Run MobSF unit tests
- [X] Tested Working on Linux, Mac,

```
could also be imporoved I think 
```


ajinabraham:
@superpoussin22 can you use `Get()` as @magaofei suggested?

superpoussin22:
will work on it for next version

PR #705, opened 2018-08-28
Apkid anti debug

superpoussin22:
### What was a problem?

```
MobSf was not detecting Anti_debug Software in apk 
```

```
search for the apkid key 'anti_debug' and siplay info if found or not 
```

- [X] Run MobSF unit tests
- [X] Tested Working on Linux, Mac

```
Need to optimize and use get method as proposed 
```

### DIFF
--- a/MalwareAnalyzer/views.py
+++ b/MalwareAnalyzer/views.py
@@ -68,6 +68,12 @@ def apkid_analysis(app_dir, apk_file):
                 "result"]["anti_disassembly"]
         else:
             apkid_dict["anti_disassembly"] = ""
+        
+        if "anti_debug" in apkid_dict["result"]:
+            apkid_dict["anti_disassembly"] = apkid_dict[
+                "result"]["anti_debug"]
+        else:
+            apkid_dict["anti_debug"] = ""
 
         if "dropper" in apkid_dict["result"]:
             apkid_dict["dropper"] = apkid_dict["result"]["dropper"]
--- a/MalwareAnalyzer/views.py
+++ b/MalwareAnalyzer/views.py
@@ -70,7 +70,7 @@ def apkid_analysis(app_dir, apk_file):
             apkid_dict["anti_disassembly"] = ""
         
         if "anti_debug" in apkid_dict["result"]:
-            apkid_dict["anti_disassembly"] = apkid_dict[
+            apkid_dict["anti_debug"] = apkid_dict[
                 "result"]["anti_debug"]
         else:
             apkid_dict["anti_debug"] = ""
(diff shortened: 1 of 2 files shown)


superpoussin22:
@magaofei can you take a look ?

templates/static_analysis/static_analysis.html
```diff
@@ -891,6 +892,17 @@ <h3 class="box-title"><i class="fa fa-bug"></i> DEX Malware Analysis</h3>
                               {% endfor %}
                           {% endif %}
                         </td>
+                       
+                        <td>
+                          {% if apkid|key:"anti_debug" == "" %}
```
magaofei: `if apkid|key:"anti_debug"|length == 0` is better ?

superpoussin22: need to dicsuss cause we need in this case to rewrite all the file, I have done a simple copy/paste

ajinabraham: I think it's fine for now to use `{% if apkid|key:"anti_debug" == "" %}`It won't break any functionality.
This file has more occurance of the same coding style. We can take care of that in a Code QA later, or do a global replace if it doesn't break anything and if it's not too much work to change and test.

MalwareAnalyzer/views.py
```diff
@@ -31,96 +31,68 @@ def apkid_analysis(app_dir, apk_file):
...
-            apkid_dict["anti_vm"] = apkid_dict["result"]["anti_vm"]
-        
-        if "compiler" in apkid_dict["result"]:
-            apkid_dict["compiler"] = apkid_dict["result"]["compiler"]
+            apkid_dict2["result"] = result["files"][1]["results"]
         
-        if "packer" in apkid_dict["result"]:
-            apkid_dict["packer"] = apkid_dict["result"]["packer"]
+        apkid_dict2['anti_vm'] = apkid_dict2['result'].get('anti_vm')  
+        if apkid_dict2['anti_vm'] is not None:
```
ajinabraham: @superpoussin22 How about`if 'anti_vm' in apkid_dict2`?

superpoussin22: @ajinabraham anti_vm should be in apkid_dict but I do a test just in case

superpoussin22: I soon as I know what is in which dictionnary (there's no spec) I will clean the code

superpoussin22: I have apdated it to make it more simpe

MalwareAnalyzer/views.py
```diff
@@ -31,84 +31,33 @@ def apkid_analysis(app_dir, apk_file):
...
-            apkid_dict["anti_disassembly"] = apkid_dict[
-                "result"]["anti_disassembly"]
-        
-        if "dropper" in apkid_dict["result"]:
-            apkid_dict["dropper"] = apkid_dict["result"]["dropper"]
-        
-        if "manipulator" in apkid_dict["result"]:
-            apkid_dict["manipulator"] = apkid_dict["result"]["manipulator"]
+        if "files" in result:
+            apkid_dict2["result"] = result["files"][1]["results"]
```
magaofei: 1 Will not get the value?

superpoussin22: I see rare case without value for position 1

superpoussin22:
all te test with diffrerent apk are ok , debugger info ar displayed ans also packer when they are present

PR #706, opened 2018-08-28
change permissons for android to dict

magaofei:
### What was a problem?

The API of the permission part of Android static analysis is not good enough. The permission field returns a dictionary. The value of the dictionary is a list instead of a dictionary, so the order must be guaranteed. This change may cause historical compatibility issues and requires caution

origin
```
{'android.permission.RECEIVE_BOOT_COMPLETED': ['normal', 'automatically start at boot', 'Allows an application to start itself as soon as the system has finished booting. This can make it take longer to start the phone and allow the application to slow down the overall phone by always running.'}
```

now
```
{'android.permission.RECEIVE_BOOT_COMPLETED': {'status': 'normal', 'info': 'automatically start at boot', 'description': 'Allows an application to start itself as soon as the system has finished booting. This can make it take longer to start the phone and allow the application to slow down the overall phone by always running.'},
```

```
DESCRIBE HERE
```

- [x] Run MobSF unit tests

```
DESCRIBE HERE
```

### DIFF
--- a/StaticAnalyzer/views/android/code_analysis.py
+++ b/StaticAnalyzer/views/android/code_analysis.py
@@ -75,9 +75,9 @@ def code_analysis(app_dir, perms, typ):
                     # print "[INFO] Doing Code Analysis on - " + jfile_path
                     relative_java_path = jfile_path.replace(java_src, '')
                     code_rule_matcher(
-                        code_findings, list(perms.keys()), dat, relative_java_path, code_rules)
+                        code_findings, perms, dat, relative_java_path, code_rules)
                     # API Check
-                    api_rule_matcher(api_findings, list(perms.keys()),
+                    api_rule_matcher(api_findings, perms,
                                      dat, relative_java_path, api_rules)
                      # Extract URLs and Emails
                     urls, urls_nf, emails_nf = url_n_email_extract(dat, relative_java_path)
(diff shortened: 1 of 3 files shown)


StaticAnalyzer/views/android/code_analysis.py
```diff
@@ -75,9 +75,9 @@ def code_analysis(app_dir, perms, typ):
                     # print "[INFO] Doing Code Analysis on - " + jfile_path
                     relative_java_path = jfile_path.replace(java_src, '')
                     code_rule_matcher(
-                        code_findings, list(perms.keys()), dat, relative_java_path, code_rules)
+                        code_findings, perms, dat, relative_java_path, code_rules)
```
ajinabraham: There are certain places the keys of the permissions are needed like
https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/StaticAnalyzer/views/shared_func.py#L366

changing it to dict will break it ?

StaticAnalyzer/views/android/code_analysis.py
```diff
@@ -75,9 +75,9 @@ def code_analysis(app_dir, perms, typ):
                     # print "[INFO] Doing Code Analysis on - " + jfile_path
                     relative_java_path = jfile_path.replace(java_src, '')
                     code_rule_matcher(
-                        code_findings, list(perms.keys()), dat, relative_java_path, code_rules)
+                        code_findings, perms, dat, relative_java_path, code_rules)
                     # API Check
-                    api_rule_matcher(api_findings, list(perms.keys()),
+                    api_rule_matcher(api_findings, perms,
```
ajinabraham: here as well

StaticAnalyzer/views/android/code_analysis.py
magaofei: This is my problem and has been restored

PR #718, opened 2018-09-06
Update apktool

superpoussin22:
### What was a problem?

```
Update apktool to 2.3.4
```

```
[#1864] Android Pie Support
[#1776] Fixed issue rebuilding applications in aapt2 with $ or _.
[#1809] Fixed process executor to wait for response when identifying aapt versions. (Thanks sunshouxiang)
[#1843] Fixed issue with 0 byte file being detected as not compressed. (Thanks alexeikh)
[#1817] Updated aapt2 builds to latest android_9.0.0_r3 build tag.
[#1889] Align codebase so aapt and aapt2 are built from same code.
[#1884] Upgrade Gradle to v5.0 to fix Java warnings.
```

- [x] Run MobSF unit tests
- [x] Tested Working on Linux, Mac, 

### DIFF
--- a/StaticAnalyzer/views/android/manifest_analysis.py
+++ b/StaticAnalyzer/views/android/manifest_analysis.py
@@ -1300,7 +1300,7 @@ def get_manifest_file(app_path, app_dir, tools_dir):
         if len(settings.APKTOOL_BINARY) > 0 and isFileExists(settings.APKTOOL_BINARY):
             apktool_path = settings.APKTOOL_BINARY
         else:
-            apktool_path = os.path.join(tools_dir, 'apktool_2.3.3.jar')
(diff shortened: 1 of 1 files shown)


magaofei:
Can we make these dynamic downloads instead of storing them in a git repository?

ajinabraham:
@magaofei There is an uncertainty in dynamic downloads. We can't relay and maintain links from third party services or app website. It may work if we host them. Unfortunately as we are non funded, we cannot reliably host binaries in the cloud like for example Amazon S3. It's a more tedious job to maintain it separately than keeping it in git.

magaofei:
@ajinabraham The git repository is so huge.

PR #733, opened 2018-09-21
Change permission to list

magaofei:
Note that it will have an impact on historical data.


ajinabraham:
@magaofei  you might need to change the PDF generating templates as well.

ajinabraham:
Why is this closed @magaofei?

magaofei:
This change is not necessary

Issue #735, opened 2018-09-22
None type Error Apk


Polibus:
## EXPLANATION OF THE ISSUE
Hi i have the problem with an android app the strange is that the other day MOBSF worked and analized the app well
I attached the log file and the link of the APK
info: I'm using the docker container on ubuntu x64 18.04 
[APK](https://play.google.com/store/apps/details?id=be.delen.digital&hl=en)
[log.log](https://github.com/MobSF/Mobile-Security-Framework-MobSF/files/2408145/log.log)

superpoussin22:
@Polibus  can you try branch: fix_encoding ? it should solve your issue

ccsplit:
@superpoussin22, there still appears to be an issue with that branch, in which `filename` is still not set. It appears [this change](https://github.com/MobSF/Mobile-Security-Framework-MobSF/commit/27aaae197ae8ec8ebddf3f78ef2d5c53fa55bff6#diff-70f9602599a659f863110d8147c81cbaL78) causes the issue since `filename = fileinfo.filename` was removed. Added this line back removes the issue for me at least.

ajinabraham:
Fixed in latest master

PR #736, opened 2018-09-23
Fix encoding

superpoussin22:
see #735   we have a lot of encoding issue

```
It solve partially a part of the error adding enconding="utf-8" to open 


```

- [x] Run MobSF unit tests

```
BTW I see a lot of encoding="utf8" in the code is it normal ? shouldn't it be utf-8 ?
```

### DIFF
--- a/StaticAnalyzer/views/shared_func.py
+++ b/StaticAnalyzer/views/shared_func.py
@@ -91,7 +91,7 @@ def unzip(app_path, ext_path):
                 subprocess.call(
                     ['unzip', '-o', '-q', app_path, '-d', ext_path])
                 dat = subprocess.check_output(['unzip', '-qq', '-l', app_path])
-                dat = dat.decode('utf-8).split('\n')
+                dat = dat.decode('utf-8').split('\n')
                 files_det = ['Length   Date   Time   Name']
                 files_det = files_det + dat
                 return files_det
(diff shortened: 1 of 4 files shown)


StaticAnalyzer/views/shared_func.py
```diff
@@ -74,12 +74,16 @@ def unzip(app_path, ext_path):
...
                 filename = fileinfo.filename
                 if not isinstance(filename, str):
                     filename = str(
                         filename, encoding="utf-8", errors="replace")
                 files.append(filename)
-                zipptr.extract(fileinfo, str(ext_path))
+                full_filename = app_path + filename
+                full_filename = bytes(full_filename, 'utf-8')
+                full_filename = str(full_filename, encoding="utf-8", errors="replace")
+                zipptr.extract(full_filename, ext_path)
```
ajinabraham: Why do you need `full_filename` . Isn't `fileinfo.filename` enough?

superpoussin22: fileinfo.filename will contain xxxxx.apk
full_filename will contain /.root/MobSF/..../.../../xxxxx.apk

this PR need nore test before merging it , i need to double check a lot of things to ensure we don't have side effect  
delen app (md5: 1eccf8b457fb615039259c002ed7a17d) give me a lot of problem so i will investigate more

superpoussin22:
@ajinabraham sounds better , only one minor error , need to works on encoding but app such delen can now be analyzed

magaofei:
If you can, remove the wide range of try catch and try to reduce the range, otherwise it will hide the real exception.

superpoussin22:
@magaofei will do it in a second step , this 1st PR should help for users with chinese code, but I thinj we can do better so it's a first quick win

PR #746, opened 2018-10-12
Remove stack protector check + small refactoring

matandobr:
As we discussed on slack...


manuumanuu:
Could you please tell me the reason for removing this check?

PR #750, opened 2018-10-16
Only warn about singleTask/singleInstance when API level is <21

vollkorn1982:
### What was a problem?

```
The underlying vulnerability is that with ActivityManager.getRecentTasks()
any app with the permission GET_TASKS can read the Intents sent to create
the task. This would reveal the possibly sensitive contents of these Intents.
MobSF warns about Activities using singleTask/singleInstance because these
would be affected by this issue.

According to the Android documentation at https://developer.android.com/reference/android/app/ActivityManager.html#getRecentTasks(int,%20int)
starting with API level 21 it is no longer possible to read all Intents sent
to a root activity by using ActivityManager.getRecentTasks(). After API
level 21, only Intents sent from your own app and other apps to be known
insensitive like the home/launcher app.

See also chapter 4.1.3.3 (page 94) in http://www.jssec.org/dl/android_securecoding_en.pdf
```

```
This PR checks if the min_sdk level is less than 21 and only then applies this issue in the manifest analysis.
```

- [x] Run MobSF unit tests

```
Only testet on Linux.

I did not find how to run the unit tests. Standard `python manage.py tests` runs 0 tests.
```

### DIFF
--- a/StaticAnalyzer/views/android/manifest_analysis.py
+++ b/StaticAnalyzer/views/android/manifest_analysis.py
@@ -762,6 +762,7 @@ def manifest_analysis(mfxml, man_data_dic):
 
                 # LaunchMode
                 if (
+                        int(man_data_dic['min_sdk']) < 21 and
                         itemname in ['Activity', 'Activity-Alias'] and
                         (
                             node.getAttribute("android:launchMode") == 'singleInstance' or
--- a/StaticAnalyzer/views/android/manifest_analysis.py
+++ b/StaticAnalyzer/views/android/manifest_analysis.py
@@ -18,6 +18,10 @@ from MobSF.utils import (
 from .dvm_permissions import DVM_PERMISSIONS
 
 
+ANDROID_4_2_LEVEL = 17
+ANDROID_5_0_LEVEL = 21
+
+
 def get_manifest(app_path, app_dir, tools_dir, typ, binary):
     """Get the manifest file."""
     try:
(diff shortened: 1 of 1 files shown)


matandobr:
It is through the browser, head to /runtest/

vollkorn1982:
```
[2018-10-16 18:43:47]
[ERROR] Performing Manifest Analysis (/home/work/Dokumente/Tools/mobsf_fork/Mobile-Security-Framework-MobSF/StaticAnalyzer/views/android/manifest_analysis.py, LINE 765 "int(man_data_dic['min_sdk']) < 21 and"): invalid literal for int() with base 10: ''
[ERROR] 'NoneType' object is not subscriptable
Internal Server Error: /StaticAnalyzer/
  [ERROR] Performing Static Analysis:  /StaticAnalyzer/?name=android_src.zip&type=zip&checksum=52c50ae824e329ba8b5b7a0f523efffe
```

I'll figure out tomorrow how to fix this.

vollkorn1982:
It's fine when running a real apk analysis. From my understanding it's impossible to build an .apk without setting a min_sdk version. Therefore, I believe this to be a missing part of the test suite. Nonetheless I could insert a test if the value does not exist or is None, just to be on the safe side.

magaofei:
on the top of file, and use this constant.
like this
```python
ANDROID_5_0_LEVEL = 21
```

vollkorn1982:
@magaofei Well, I built this in the fashion of the already existing code at line 1034 and 1035 and would make code clean-ups in a separate PR. But here you go.

vollkorn1982:
Btw: You really shouldn't talk about constants in Python. There is no such thing. (At least in the Python language or stdlib. You can build your own consts though.)

magaofei:
@vollkorn1982 https://www.python.org/dev/peps/pep-0008/#constants

vollkorn1982:
@magaofei From the category "Naming Conventions". You see the problem yourself? ;)

Anymore changes? Please don't feed them to me one by one, but make a list I can work off at once. Thank you.

magaofei:
@vollkorn1982 If you come over and discuss the constants and naming conventions with me, you don't need them. I only pay attention to the results.

PR #755, opened 2018-10-24
Remove ios strings

### LINKED ISSUE #754: It is recommended to remove the strings in the iOS api response results.
Upload an ipa file on the web page and the scan will take only a few seconds. Using the same file request API takes about 2 minutes. I guess it is because the API contains strings information. I suggest removing it or using a separate API.

- When you use the template to open the scan results, it does not directly display the strings information, a button will be placed to click, so strings are not displayed by default. But when I go to the API, the results of the API will be taken with strings.
- API response time is too long

magaofei:
fix #754 

Unfortunately, ipa can't decompile useful source code information like apk, I recommend removing it.

### DIFF
--- a/StaticAnalyzer/views/ios/binary_analysis.py
+++ b/StaticAnalyzer/views/ios/binary_analysis.py
@@ -301,7 +301,6 @@ def binary_analysis(src, tools_dir, app_dir, executable_name):
         bin_path = os.path.join(bin_dir, bin_name)
         binary_analysis_dict["libs"] = ''
         binary_analysis_dict["bin_res"] = ''
-        binary_analysis_dict["strings"] = ''
         if not isFileExists(bin_path):
             print("[WARNING] MobSF Cannot find binary in " + bin_path)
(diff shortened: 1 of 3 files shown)


ajinabraham:
@magaofei As discussed over slack. This is an essential feature for malware analysts. So we can't remove it. If the API is slower than the corresponding Web UI. Then we need to investigate that and fix it rather than removing a feature.

magaofei:
@ajinabraham In another way, don't delete it. Instead, create a separate table to store and create a new API access

ajinabraham:
Having multiple API to retrieve the results of an IPA is not neat. How about we modify strings to reduce the amount of data extracted.

magaofei:
In my case, ipa's strings are mostly confusing, and they are useless, and the number is particularly large. Rules may need to be modified

ajinabraham:
We can always optimize the strings function, but can't remove it. It has quite a lot of useful information from malware analysis perspective.

ajinabraham:
@magaofei Can I close this?

PR #757, opened 2018-10-26
REST API QA

ajinabraham:
### What was a problem?

```
Code Quality
```

```
* View Source API Code QA
* REST API for view source simplified
* REST API Docs
* View Source Unit Tests
* General Code QA
```

- [x] Run MobSF unit tests

```
DESCRIBE HERE
```


StaticAnalyzer/views/shared_func.py
```diff
@@ -204,31 +204,32 @@ def pdf(request, api=False):
...
-                    ],
-                    'no-outline': None
-                }
-                pdf_dat = pdfkit.from_string(html, False, options=options)
-                if api:
-                    return {"pdf_dat": pdf_dat, "report_dat": context}
+                if api and json:
+                    return {"report_dat": context}
                 else:
+                    options = {
```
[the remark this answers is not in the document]
ajinabraham: What you mean?
That view generates JSON, PDF Raw and PDF HTTP Response.

MobSF/views/api/rest_api.py
```diff
@@ -181,32 +179,22 @@ def api_json_report(request):
...
-    viewsource_form = ViewSourceIosForm(request.GET)
-    if not viewsource_form.is_valid():
-        return JsonResponse(FormUtil.errors_message(viewsource_form), status=BAD_REQUEST)
-
-    view_source = ViewSourceIos(request)
-    return view_source.api()
-    
-
+    params = ['file', 'type', 'hash']
+    if set(request.POST) >= set(params):
```
magaofei: It seems that there is no need to check the parameters here.

ajinabraham: You are right. But because we have form checks in the other function. But since other functions use the same pattern and they don't have form validation. I thought to keep it as this should not confuse anyone in future.

I will remove that probably later when we have form checks for every API. We will do that as a part of code QA.

PR #792, opened 2018-11-29
support deleted not found tips

magaofei:
If it does not exist, it should return a non-existent state

### DIFF
--- a/MobSF/views/home.py
+++ b/MobSF/views/home.py
@@ -109,7 +109,8 @@ class Upload(object):
         if self.file_type.is_ipa():
             if platform.system() not in LINUX_PLATFORM:
                 print("\n[ERROR] Static Analysis of iOS IPA requires Mac or Linux")
-                response_data['description'] = 'Static Analysis of iOS IPA requires Mac or Linux'
+                response_data[
+                    'description'] = 'Static Analysis of iOS IPA requires Mac or Linux'
                 response_data['status'] = 'success'
                 response_data['url'] = 'mac_only/'
                 return self.resp_json(response_data)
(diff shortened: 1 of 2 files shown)


MobSF/views/home.py
```diff
@@ -298,6 +298,9 @@ def delete_scan(request, api=False):
                         if isDirExists(item_path) and item.startswith(md5_hash + "-"):
                             shutil.rmtree(item_path)
                     data = {'deleted': 'yes'}
+                else:
+                    if api:
+                        return {'deleted': 'not_found'}
```
ajinabraham: Is this required?
Why not provide it here: https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/MobSF/views/home.py#L276

You might need to change the JS Logic in template as well

magaofei: In theory, it should return a 404 status code. When I call this API, I don't know if this is not deleted because the error still doesn't exist, because I may have deleted it through the page, so I need to distinguish some states.

ajinabraham: Hmf. Let me test the behaviour once.

ajinabraham: @magaofei I am seeing the JSON response message for already deleted items

```
curl -X POST --url http://localhost:8000/api/v1/delete_scan --data "hash=3a552566097a8de588b8184b059b0158" -H "Authorization:***"
{"deleted": "no"}
```

ajinabraham: What is that you see as an issue here?

magaofei: When deleting a project that does not exist, it should return a state that does not exist, which means that there is no such project, not the deletion error.

ajinabraham:
Now the API will give more informative response like
```
{"deleted": "scan hash not found"}
```

PR #793, opened 2018-11-29
fix viewsource for plist file format

magaofei:
Plist file will be garbled if it is read directly

### DIFF
--- a/StaticAnalyzer/views/ios/view_source.py
+++ b/StaticAnalyzer/views/ios/view_source.py
@@ -5,6 +5,7 @@ iOS View Source
 import re
 import os
 import io
+import json
 import ntpath
 import sqlite3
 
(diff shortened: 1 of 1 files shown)


ajinabraham:
@magaofei what happens if the plist file is not a binary plist?

StaticAnalyzer/views/ios/view_source.py
```diff
@@ -34,7 +35,9 @@ def set_ext_api(file_path):
     """
     ext = file_path.split('.')[-1]
     if ext == "plist":
-        return "xml"
+        return "plist"
```
ajinabraham: :TODO : check if `xml` has to do something with Syntax highlighting.

StaticAnalyzer/views/ios/view_source.py
```diff
@@ -84,6 +87,9 @@ def run(request, api=False):
             file_format = 'xml'
             with io.open(sfile, mode='r', encoding="utf8", errors="ignore") as flip:
                 dat = flip.read()
+        elif typ == 'plist':
+            file_format = 'plist'
+            dat = biplist.readPlist(sfile)
```
ajinabraham: What happens if it reads a non binary plist. Like from an iOS Source Zip

magaofei:
Just like Android's Manifest will be specially processed, iOS plist files are also needed, and can not be read directly like xml files. In my tests, some files can be read directly, but more files will be garbled.

PR #794, opened 2018-11-30
more skip classes + cfr silent mode

matandobr:
- added classes - okhttp2/3, android.arch, kotlin, 
- cfr in silent mode, the terminal output is not good for any reason...


ajinabraham:
@matandobr Why is the diff so weird? Any line ending changes or something?

matandobr:
@ajinabraham 
Yeah I guess you are right, I'll fix it

PR #797, opened 2018-12-04
Move to JAVA 11 in Docker File

superpoussin22:
### What was a problem?

```
Java 11 is now the LTS version for Java , so the dockerfile need to be updated
```

```
It add Java 11 in the image
```

- [X] Run MobSF unit tests (http://your-mobsf-@ip:8000/tests/ or python3 manage.py test)
- [X] Tested Working on Linux, Mac, and Windows and Docker

```
None
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,9 +91,6 @@ RUN git clone --recursive ${YARA_URL} yara-python && \
     rm -fr /tmp/yara-python && \
     sed -i 's/APKID_ENABLED.*/APKID_ENABLED = True/' /root/Mobile-Security-Framework-MobSF/MobSF/settings.py
 
-# Set Java 11 Directory
-RUN sed -i 's#JAVA_DIRECTORY = ""#JAVA_DIRECTORY = "/usr/bin/"#' /root/Mobile-Security-Framework-MobSF/MobSF/settings.py
-
 #Cleanup
 RUN \
     apt remove -y git && \
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,12 +29,19 @@ RUN apt update -y && apt install -y \
     libxml2-dev \
     libxslt1-dev
 
-#Install Oracle JDK 8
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -29,12 +29,19 @@ RUN apt update -y && apt install -y \
     libxml2-dev \
     libxslt1-dev
 
-#Install Oracle JDK 8
```
ajinabraham: Please remove unnecessary comments

Dockerfile
```diff
@@ -91,6 +98,9 @@ RUN git clone --recursive ${YARA_URL} yara-python && \
     rm -fr /tmp/yara-python && \
     sed -i 's/APKID_ENABLED.*/APKID_ENABLED = True/' /root/Mobile-Security-Framework-MobSF/MobSF/settings.py
 
+# Set Java 11 Directory
+RUN sed -i 's#JAVA_DIRECTORY = ""#JAVA_DIRECTORY = "/usr/bin/"#' /root/Mobile-Security-Framework-MobSF/MobSF/settings.py
```
ajinabraham: Is this required. Wouldn't Java 11 be the default. Instead of this. Make sure Java 11 is the default.

ajinabraham: You should investigate why it's failing.

Dockerfile
superpoussin22: done

Dockerfile
superpoussin22: it come's from the fact that script in profile.d are not launched , so the jdk.sh script which set the ENV is not sourced , shell in docker is not interactive

ajinabraham: @superpoussin22 Can we avoid setting Java directory as in line 102?

superpoussin22: @ajinabraham yep done 
I modify utils.py cause /usr/bin/python did note provide Oracle in output 
let me know if ok for you

PR #799, opened 2018-12-06
remove configparser

noraj:
### What was a problem?

```
This library brings the updated configparser from Python 3.5 to Python 2.6-3.5.
```

So unneeded if using python3, I guess it was when mobsf was in python2.

removes the unneeded dependency

- [ ] Run MobSF unit tests (http://your-mobsf-@ip:8000/tests/ or python3 manage.py test)

I didn't tested because I don't have a proper setup and I don't want to trash my actual machine by installing last version of mobsf.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 Django==2.1.4
 pdfkit==0.6.1
-configparser==3.5.0
 androguard==3.2.1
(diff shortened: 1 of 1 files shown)


noraj:
@ajinabraham so it wasn't needed anymore ?

PR #802, opened 2018-12-07
Update clean.sh

superpoussin22:
prevent script from being launched outside of Mobsf path

### What was a problem?

```
clean script could delete directory outside of MobSF path if launched outside of MobSF path
```

```
it check it script is launched from .  or from ./scripts
```

- [ ] Run MobSF unit tests (http://your-mobsf-@ip:8000/tests/ or python3 manage.py test)
- [X] Tested Working on Linux, Mac, and Windows

```

```

### DIFF
--- a/scripts/clean.sh
+++ b/scripts/clean.sh
@@ -3,7 +3,7 @@ echo "=======================MobSF Clean Script for Unix======================="
 echo "Running this script will delete the Scan database, all files uploaded and generated."
 
 script_path=$(dirname $0)
-if [ "$script_path" != "." ] && [ "$script_path" != "./scripts" ]; then
+if [ "$script_path" != "." ] && [ "$script_path" != "scripts" ] && [ "$script_path" != "./scripts" ]; then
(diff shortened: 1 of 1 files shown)


superpoussin22:
add better support for setup.sh

scripts/clean.sh
```diff
@@ -2,6 +2,12 @@ echo
 echo "=======================MobSF Clean Script for Unix======================="
 echo "Running this script will delete the Scan database, all files uploaded and generated."
 
+script_path=$(dirname $0)
+if [ "$script_path" != "." ] && [ "$script_path" != "scripts" ] && [ "$script_path" != "./scripts" ]; then
```
ajinabraham: Will this affect `setup.sh` ?
https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/setup.sh#L21

superpoussin22: setup.sh will continue working  it's the case where script_path="scripts"

PR #806, opened 2018-12-12
manifest add key

magaofei:
### What was a problem?

```
The data structure is more reasonable, and new fields have been added to indicate which problem
```

```
DESCRIBE HERE
```

- [x] Run MobSF unit tests (http://your-mobsf-@ip:8000/tests/ or python3 manage.py test)

```
DESCRIBE HERE
```


magaofei:
I want to get the permission name through this API, but I found that the original title comes with some extra content, such as the name of the activity, I want a fixed name to distinguish.

ajinabraham:
It's fine but In my opinion, that should ideally be done at the client side who ever is using the API.

magaofei:
Yes, this content will be returned to the people who use the API so they can easily distinguish permissions

PR #808, opened 2018-12-13
change request method decorator param to str

magaofei:
### What was a problem?

```
Request method does not need to support multiple
```

```
DESCRIBE HERE
```

- [x] Run MobSF unit tests (http://your-mobsf-@ip:8000/tests/ or python3 manage.py test)

```
DESCRIBE HERE
```

### DIFF
--- a/MobSF/views/api/rest_api.py
+++ b/MobSF/views/api/rest_api.py
@@ -59,7 +59,7 @@ def api_auth(meta):
     return False
 
 
-@request_method(['POST'])
+@request_method('POST')
 @csrf_exempt
 def api_upload(request):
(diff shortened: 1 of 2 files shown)


ajinabraham:
@magaofei Why string?
In future if we want to support multiple HTTP verbs to a single API string won't be sufficient.

magaofei:
support multiple HTTP verbs ? Doesn't seem to need it?

ajinabraham:
@magaofei Not now, but in future if we need. Then tuples or lists are useful to support multiple verbs. What is the problem you see with lists now?

magaofei:
I have no problem, we can keep it.

PR #812, opened 2018-12-14
Android - APP compare

matandobr:
- Currently only supports android
- All tests passed without errors
- The recent.html changed because of the the diff selection
- The urls.py changed becuase there was a bug in complicated urls, so I used named urls in the base.html
- All the compare logic & new html...


StaticAnalyzer/views/comparer.py
```diff
@@ -0,0 +1,128 @@
+# -*- coding: utf_8 -*-
+"""
+Module providing the shared functions for static analysis of iOS and Android
+"""
+import re
+from copy import deepcopy
+
+try:
+    import pdfkit
```
ajinabraham: Is this needed?

StaticAnalyzer/views/shared_func.py
```diff
@@ -528,3 +529,12 @@ def url_n_email_extract(dat, relative_path):
         email_n_file.append(
             {"emails": emails, "path": escape(relative_path)})
     return urllist, url_n_file, email_n_file
+
+
+# This is just the first sanity check that triggers generic_compare
+def compare_apps(request, first_hash: str, second_hash: str):
+    if first_hash == second_hash:
+        error_msg = "2 same hashes were provide for comparison"
```
ajinabraham: Can you reword this to `Results with same hashes cannot be compared` ?

templates/general/recent.html
```diff
@@ -1,92 +1,204 @@
...
+
+    function disable_diff_button() {
+        $('.diffButton').off('click');
+    }
+
+    // First pop up only saves the first scan to diff and tells the user to select a partner
+    function diff_select(item) {
+
+        swal({
+            title: '<strong>Select partner</strong>',
```
ajinabraham: Would be nice to title it as `Select App to Compare`

templates/general/recent.html
```diff
@@ -1,92 +1,204 @@
...
+            type: 'info',
+            html:
+                'Please select which scan to diff with',
```
ajinabraham: `Please select the second scan result for comparison`

ajinabraham:
We should show an understandable error message when diffing non apk results. 

```
Error Description

One of the Hashes was not found in the DB
```

ajinabraham:
@matandobr Tested the functionality. It's brilliant and quite useful!
Please see minor comments.

matandobr:
@ajinabraham 
I'm really glad you like it :)
fixed all the issues.

PR #1009, opened 2019-07-20
Fix malware false positive

superpoussin22:
### Describe the Pull Request

```
using in instead of startswith will produce false positive example for chrombeta which use 
g.co. if we use in it will match malware url 
```

### Checklist for PR


superpoussin22:
I suppress the length test on line 93 , to be compatible with E501 the only solution will be to reduce variable length :(  or change the way of testing

PR #1045, opened 2019-08-08
Fix Android insecure Random Number Generator static rule

peja:
This fixes an issue with "java.util.RandomAccess" being flagged by mistake because it contains "java.util.Random".


peja:
> Nice, probably we can use `\b` with the iOS otool use case as well..

I'm about to try it now :)

PR #1052, opened 2019-08-08
Improve iOS static analyzer accuracy

peja:
Static analyzer would previously match only a part of the function name, which could lead to false positives in some cases, e.g. **_objc_allocateProtocol** function would be flagged as **_alloca**.


ajinabraham:
closing in favour of https://github.com/MobSF/Mobile-Security-Framework-MobSF/pull/1054

PR #1059, opened 2019-08-10
Reduce Docker image size by replacing wkhtml with static compiled 

superpoussin22:
### Describe the Pull Request

```

- Reduce Docker image Size 

- Reorder command to reduce layer

- Reduce Layer

- Silent some command

- Use Static compiled version of Wkhtmltopdf 

- Remove no more needed binaries

```

### Checklist for PR

```
wkhtml static version is downloaded from internet but if needed we can put it in MobSF repo
Docker file is reordered to have a base layer in case we decide to split it 

```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -12,9 +12,11 @@ LABEL \
 
 #Environment vars
 ENV DEBIAN_FRONTEND="noninteractive" \
-    JDK_FILE="openjdk-12_linux-x64_bin.tar.gz"
+    JDK_FILE="openjdk-12_linux-x64_bin.tar.gz" \
+    WKH_FILE="wkhtmltox-0.12.5-dev-163e124_linux-generic-amd64.tar.xz"
 
-ENV JDK_URL="https://download.java.net/java/GA/jdk12/GPL/${JDK_FILE}" 
+ENV JDK_URL="https://download.java.net/java/GA/jdk12/GPL/${JDK_FILE}" \
+    WKH_URL="http://www.ajvg.com/downloads/${WKH_FILE}"
 
 #Update the repository sources list
 #Install Required Libs
@@ -44,6 +46,11 @@ RUN apt update -y && apt install -y \
 RUN locale-gen en_US.UTF-8
 ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
 
+#Install Wkhtmltopdf
+RUN wget --quiet -O /tmp/wkhtmltox.tar.xz "${WKH_URL}" && \
+    tar xJf /tmp/wkhtmltox.tar.xz -C /usr/bin/ --strip-component=2 wkhtmltox/bin/wkhtmltopdf && \
+    rm -f /tmp/wkhtmltox.tar.xz
+
 #Install OpenJDK12
 RUN wget --quiet "${JDK_URL}" && \
     tar zxf "${JDK_FILE}" && \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -11,61 +11,59 @@ LABEL \
...
+    libxrender1 \
+    libfontconfig1 \ 
+    libxext6 \
+    fontconfig \
+    xfonts-75dpi \
+    python3.6 \
+    python3-dev \
+    python3-setuptools \
+    wget && \
+    python3 /usr/lib/python3/dist-packages/easy_install.py pip
```
ajinabraham: There is no other documented way to install python3-pip?

ajinabraham: wouldn't `pip3` already be available?

superpoussin22:
> Do we need this if we can move to alpine?

For the moment "mac tools" are not working but I'm on it , so no alpine build

Dockerfile
superpoussin22: will change that, you're right

superpoussin22: pip is not installed by default

Dockerfile
```diff
@@ -11,61 +11,58 @@ LABEL \
...
 
 #Install wkhtmltopdf for PDF Reports
 WORKDIR /tmp
-RUN wget --quiet ${PDFGEN_URL} && \
-    dpkg -i ${PDFGEN_PKGFILE} && \
-    rm -rf ${PDFGEN_PKGFILE}
+RUN wget --quiet "${WKH_URL}" && \
+    tar xJf "${WKH_FILE}"
+WORKDIR /tmp/wkhtmltox/bin
+RUN cp ./* /usr/bin
```
ajinabraham: Why not symlink and reduce a copy

ajinabraham: @superpoussin22

ajinabraham:
@superpoussin22 also see Codacy PR quality

Dockerfile
```diff
@@ -11,61 +11,58 @@ LABEL \
...
-ENV PDFGEN_PKGFILE="wkhtmltox_0.12.5-1.bionic_amd64.deb" 
-ENV PDFGEN_URL="https://github.com/wkhtmltopdf/wkhtmltopdf/releases/download/0.12.5/${PDFGEN_PKGFILE}"
-ENV JDK_FILE="openjdk-12_linux-x64_bin.tar.gz"
-ENV JDK_URL="https://download.java.net/java/GA/jdk12/GPL/${JDK_FILE}"
+ENV DEBIAN_FRONTEND="noninteractive" \
+    JDK_FILE="openjdk-12_linux-x64_bin.tar.gz" \
+    WKH_FILE="wkhtmltox-0.12.5-dev-163e124_linux-generic-amd64.tar.xz"
+
+ENV JDK_URL="https://download.java.net/java/GA/jdk12/GPL/${JDK_FILE}" \
+    WKH_URL="http://www.ajvg.com/downloads/${WKH_FILE}"
```
ajinabraham: Which is this source?
Is it official?

superpoussin22: that's a patched version made to reduce wkhtml problem , it will be the same if we switch to  Alpine we will need to use a patched version .
We can stay on official version, on this case I will back port Dockerfile optimisation to the version with official wkhtmltopdf  but we won't see big improvement in size for the image

superpoussin22: not an official but static compiled and patched regarding QT5

ajinabraham: That's okey, but we stick to official builds as much as possible.

superpoussin22: will see if I can symlink instead of making copy

Dockerfile
```diff
@@ -53,20 +51,14 @@ RUN wget --quiet "${JDK_URL}" && \
...
-    rm -f "${WKH_FILE}"
-
 #Add MobSF master
 COPY . /root/Mobile-Security-Framework-MobSF
 WORKDIR /root/Mobile-Security-Framework-MobSF
 
+#Symlink wkhtml tools
+RUN ln -s /root/Mobile-Security-Framework-MobSF/StaticAnalyzer/tools/wkhtmltox/bin/wkhtmltopdf /usr/bin/ \
+    ln -s /root/Mobile-Security-Framework-MobSF/StaticAnalyzer/tools/wkhtmltox/bin/wkhtmltoimage /usr/bin/
+
```
ajinabraham: @superpoussin22 Is line 60 needed. We only want wkhtmltopdf?

Also why not 

```
RUN ln -s /root/Mobile-Security-Framework-MobSF/StaticAnalyzer/tools/wkhtmltox/bin/wkhtmltopdf /usr/bin/wkhtmltopdf
```

superpoussin22: will remove the unneeded binary

superpoussin22:
> @superpoussin22 we don't want to bundle binaries with the project for the sake of Dockerfile or Linux

ok will reintroduce download :)

PR #1100, opened 2019-09-30
Fixed bug in the iOS static analyzer when the binary name was not the…

AdriVillaB:
… same as the package name

### Describe the Pull Request

```
The iOS Static Analyzer had a bug when the binary name was not the same as the package name. On the if clause, the is_file_exists function received just the binary name, without the complete path, thus this function returns always false.
```

### Checklist for PR


ajinabraham:
Nice Catch!

PR #1149, opened 2019-10-30
Update setup.sh

MRazian:
### Describe the Pull Request

```
MobSF requires Python3.6 or Python3.7 to work properly in dynamic analysis mode. However, in the setup.sh file, the python used for creation of virtual environment is python3. So, we changed python3 to python3.7. We checked and now dynamic analysis works with Genymotion.
```


ajinabraham:
@MRazian MobSF doesn't have any issues with Python3.8. Some dependency used by MobSF like frida was not Python 3.8 complaint. They will soon start supporting it. This fix will cause issue for Python 3.6 users.

PR #1167, opened 2019-11-11
Improvements to MobSF Docker installation

feeltheajf:
### Describe the Pull Request

```
- Fixed typo in docker-compose.yml
- Updated Dockerfile to install requirements before copying source code. This one dramatically decreases image rebuild time
- Added option to override secret key using environment variable. As of now, secret file is created upon image build. As a result, secret is the same for all installations
```

### Checklist for PR

```
DESCRIBE HERE
```

### DIFF
--- a/MobSF/utils.py
+++ b/MobSF/utils.py
@@ -33,6 +33,7 @@ from . import settings
 logger = logging.getLogger(__name__)
 ADB_PATH = None
 
+
 class Color(object):
     GREEN = '\033[92m'
     ORANGE = '\033[33m'
--- a/MobSF/utils.py
+++ b/MobSF/utils.py
@@ -33,7 +33,6 @@ from . import settings
 logger = logging.getLogger(__name__)
 ADB_PATH = None
 
-
 class Color(object):
     GREEN = '\033[92m'
     ORANGE = '\033[33m'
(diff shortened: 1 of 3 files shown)


MobSF/utils.py
```diff
@@ -503,8 +503,9 @@ def check_basic_env():
 
 def first_run(secret_file, base_dir, mobsf_home):
     # Based on https://gist.github.com/ndarville/3452907#file-secret-key-gen-py
-
-    if is_file_exists(secret_file):
+    if 'MOBSF_SECRET_KEY' in os.environ:
```
ajinabraham: Why do you need a static secret? If you just need a predefined REST `API_KEY` , use this env variable instead.

https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/f56135bcc751c451142de5d8b8947690ea5d5b46/MobSF/utils.py#L72

feeltheajf: Sorry, I don't really understand. What do you mean by 'static secret'? It is static right now because secret file is generated when the image is built, not on container startup. As a result, API_KEY is always the same by default as well. What I suggest is an option to set secret key on container run using environment variable

ajinabraham: I got confused, yes you are right.

PR #1193, opened 2019-12-06
Proxy, ADB and frida improvements

toxicphreAK:
# Describe the Pull Request

* The pull request adds improvements of the CA certificate, as the file conversion was wrong. 
* Additionally, the Proxy IP is not dynamic via settings, instead of a hardcoded localhost. 
* Starting ADB as root as it is needed in some cases.
* Improve frida process identification.

### Checklist for PR


DynamicAnalyzer/views/android/environment.py
```diff
@@ -95,17 +113,26 @@ def configure_proxy(self, project):
 
     def install_mobsf_ca(self, action):
         """Install or Remove MobSF Root CA."""
+        ca_construct = '{}.0'
+        ca_file_hash = subprocess.check_output(['openssl',
```
ajinabraham: Could you please explain why this is required?

DynamicAnalyzer/views/android/environment.py
```diff
@@ -46,10 +46,28 @@ def connect_n_mount(self):
...
+        self.wait(2)
         if b'unable to connect' in out or b'failed to connect' in out:
             logger.error('%s', out.decode('utf-8').replace('\n', ''))
             return False
         else:
+            # TODO: proof if not already connected as admin 
+            # and not already FS is mounted
+            # start adb as root
+            logger.info('Restarting ADB Daemon as root')
+            out = subprocess.check_output([get_adb(), 'root'])
```
ajinabraham: Genymotion by default has root user in shell. Why is this required?

DynamicAnalyzer/views/android/environment.py
```diff
@@ -115,7 +142,7 @@ def set_global_proxy(self, version):
         if version < 5:
             proxy_ip = get_proxy_ip(self.identifier)
         else:
-            proxy_ip = '127.0.0.1'
+            proxy_ip = settings.PROXY_IP
```
ajinabraham: The Proxy IP is doesn't have to be dynamic for android version >=5.
We make use of adb reverse tcp (https://blog.grio.com/2015/07/android-tip-adb-reverse.html) to make proxy available to device via localhost. So it is static.

DynamicAnalyzer/views/android/environment.py
```diff
@@ -405,7 +432,7 @@ def frida_setup(self):
     def run_frida_server(self):
         """Start Frida Server."""
         check = self.adb_command(['ps'], True)
-        if b'/system/fd_server' in check:
+        if b'fd_server' in check:
```
ajinabraham: was `/system/` not available for `ps` ?

DynamicAnalyzer/views/android/environment.py
toxicphreAK: The mitmproxy script is generating a CA cert at first start. The conversation to store the certificate in the Android certificate store is to use a part of the hash. this is build with the constuctor in the code and the has, identified with openssl. (originally it was hardcoded in the config)

DynamicAnalyzer/views/android/environment.py
toxicphreAK: Yes, you are right - but in case somebody is using another "device" it is required and since it does not has any negative effect on genymotion it would be nice to eventually support other setups. (I am hosting my whole setup as a docker-compose script - if you are interested in, I could share it, after I finis my studies with your nice tool!)

DynamicAnalyzer/views/android/environment.py
toxicphreAK: You are right with your setup, but as I am using docker, the container is attached to a dedicated docker network, not having mobfs and the android emulator on the same host. So making it default to localhost in the config would be a default and changeable to other setups.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: for me the output didn't work with the system in the string. It was just "system" in my output. And as this is just more lax, I think the precision of finding only "fd_server" would be nearly the same, since I don't now any other similar name by default on andoird.

DynamicAnalyzer/views/android/environment.py
ajinabraham: So does the name of cert matters to Android to identify the user installed certificate?

If so, there needs to be an alternative to openssl. This won't work in default windows setup.

DynamicAnalyzer/views/android/environment.py
ajinabraham: That would be interesting. We were testing docker support, the issues we faced were getting proxy to work and some adb conflict if there is an adb server running in host.

DynamicAnalyzer/views/android/environment.py
ajinabraham: So currently this setup won't work with non x86 VMs, device or emulators. This require changes to the frida binary.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: I understand the problem on windows - but I didn't get it as I am running on linux / Ubuntu Docker. Additionally, I cannot say if this is required, but this is the way normally be conform with the storage of a CA certificate file. You have to decide, as it is just a suggestion. Sorry that I have no further information of how to handle it on windows.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: I created a gist for you to have a look, as it is my actual setup (running on a Kali Linux VM on VMware Workstation 15 Pro on a Windows 10 host system). I you have further questions we could get in contact! :)
See: https://gist.github.com/toxicphreAK/8ecfaecb76c69900e1bb95f52324d5af

toxicphreAK:
Then the additional architectures have to be included into the repository and after the architecture of the Android Virtual Device is detetced the right version has to be copied to the device.

ajinabraham:
Python client also need update: https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/requirements.txt#L15

DynamicAnalyzer/views/android/environment.py
superpoussin22: for openssl you should use shutil.which() to find the full path of openssl and if openssl is available in the path and raise an error if not

DynamicAnalyzer/views/android/environment.py
ajinabraham: This looks interesting. I will definitely take a look after my vacation.

DynamicAnalyzer/views/android/environment.py
ajinabraham: We can't have it in code if it's not cross platform without external dependencies. If the naming doesn't affect the functionality, I suggest to keep it hardcoded like before.

Or may be something like

```
>>> import os
>>> from OpenSSL import crypto
>>> pem = open(os.path.expanduser('~/.mitmproxy/mitmproxy-ca-cert.pem'), 'rb').read()
>>> cert=crypto.load_certificate(crypto.FILETYPE_PEM, pem)
>>> cert.subject_name_hash()
2344488589
```

Not sure if it's the same as `subject_hash_old`

DynamicAnalyzer/views/android/environment.py
```diff
@@ -420,4 +455,6 @@ def start_frida():
         trd = threading.Thread(target=start_frida)
         trd.daemon = True
         trd.start()
-        logger.info('Frida Server is running')
+        logger.info('Frida Server is going to start')
```
ajinabraham: Can you reword this to `Starting Frida Server`

MobSF/settings.py
```diff
@@ -312,8 +312,8 @@
     # ==============================================
 
     # ================HTTPS PROXY ===============
+    PROXY_IP = '127.0.0.1'
     PROXY_PORT = 1337  # Proxy Port
-    ROOT_CA = '0025aabb.0'
```
ajinabraham: If we don't have a programatic cross platform way to generate this from PEM file without depending on 3rd party installation locations, Let's hardcode this value. To be honest, we should generate the cert name only if the hardcoded certificate name is not respected.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: Thank you for the good alternative solution. I am trying to apply your suggestion (as it is not the same as in openssl).
Background information, why I think it would be useful (for performance):

OpenSSL has implemented a filename technique to reduce the Certificate Authority (CA) look-up time. Without this approach it would be required to read all CA files in the CApath to find the matching certificate.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: Changed it, please review again.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: The only thing missing was the conversion to hex. Now it is using the newer SHA1 hash than the "old" MD5 one. Should be fine, as it is tested in my setup. Please review.

MobSF/settings.py
toxicphreAK: Can be removed if the above change is accepted.

toxicphreAK:
Is there something wrong - or what can I do/fix that you merge the pull request? Or should I close it?

superpoussin22:
all the team is on vacation, stay tuned :)

ajinabraham:
@toxicphreAK I will re-review once I am back after vacation.

DynamicAnalyzer/views/android/environment.py
```diff
@@ -95,17 +117,26 @@ def configure_proxy(self, project):
 
     def install_mobsf_ca(self, action):
         """Install or Remove MobSF Root CA."""
+        ca_construct = '{}.0'
+        pem = open(get_ca_dir(), 'rb').read()
+        ca_file = crypto.load_certificate(crypto.FILETYPE_PEM, pem)
+        ca_file_hash = hex(ca_file.subject_name_hash()).lstrip('0x')
```
ajinabraham: nice!

DynamicAnalyzer/views/android/environment.py
```diff
@@ -95,17 +117,26 @@ def configure_proxy(self, project):
...
+                              ca_file])
             self.adb_command(['chmod',
                               '644',
                               ca_file], True)
         elif action == 'remove':
             logger.info('Removing MobSF RootCA')
-            self.adb_command(['rm', ca_file], True)
+            self.adb_command(['rm',
+                              ca_file], True)
+        # TODO: determine if an 'adb restart' is needed
```
ajinabraham: Remove this comment if not required

toxicphreAK: Not required any more, will be addressed in the next commits.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: Thanks for your support!

DynamicAnalyzer/views/android/environment.py
```diff
@@ -46,10 +48,30 @@ def connect_n_mount(self):
...
+            if b'unable to connect' in out or b'failed to connect' in out:
+                logger.error('%s', out.decode('utf-8').replace('\n', ''))
+                return False
+            logger.info('Reconnect to Android Device')
+            # connect again with root adb
+            out = subprocess.check_output([get_adb(),
+                                          'connect',
+                                           self.identifier])
+            self.wait(2)
+            if b'unable to connect' in out or b'failed to connect' in out:
```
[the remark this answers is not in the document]
toxicphreAK: Made two new functions. Can you please verify and say if you know an improvement for "routing" the True/False return values from other functions to the function who called it?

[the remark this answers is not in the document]
ajinabraham: Sorry I did not understand you. If you can make the commits, I can comment on it.

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
+        if b'unable to connect' in output or b'failed to connect' in output:
+            logger.error('%s', output.decode('utf-8').replace('\n', ''))
+            return False
+        else:
+            return True
+
+    def run_subprocess_verify_output(self, command):
+        out = subprocess.check_output(command)
+        self.wait(2)
+        if not proof_connect(out):
```
ajinabraham: `self. proof_connect` ?

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
         self.adb_command(['kill-server'])
         self.adb_command(['start-server'])
         logger.info('ADB Restarted')
         self.wait(2)
         logger.info('Connecting to Android %s', self.identifier)
-        out = subprocess.check_output([get_adb(), 'connect', self.identifier])
-        self.wait(2)
-        if b'unable to connect' in out or b'failed to connect' in out:
-            logger.error('%s', out.decode('utf-8').replace('\n', ''))
+        if not run_subprocess_verify_output([get_adb(),
```
ajinabraham: `self. run_subprocess_verify_output` ?

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
-                return False
-            # mount system
-            logger.info('Remounting /system')
-            self.adb_command(['mount', '-o',
-                              'rw,remount', '/system'], True)
+        # TODO: proof if not already connected as admin
+        # and not already FS is mounted
+        # start adb as root
+        logger.info('Restarting ADB Daemon as root')
+        if not run_subprocess_verify_output([get_adb(), 'root']):
```
ajinabraham: missing self

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
+            return False
+        logger.info('Reconnect to Android Device')
+        # connect again with root adb
+        if not run_subprocess_verify_output([get_adb(),
```
ajinabraham: missing self.

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
         logger.info('Waiting for %s seconds...', str(sec))
         time.sleep(sec)
 
+    def proof_connect(self, output):
```
ajinabraham: Rename function name to `check_connect_error`

DynamicAnalyzer/views/android/environment.py
ajinabraham: neat way is to `return self.proof_connect(out)`

DynamicAnalyzer/views/android/environment.py
```diff
@@ -135,7 +143,6 @@ def install_mobsf_ca(self, action):
             logger.info('Removing MobSF RootCA')
             self.adb_command(['rm',
                               ca_file], True)
-        # TODO: determine if an 'adb restart' is needed
         # with a high timeout afterwards
```
ajinabraham: remove this comment as well

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
         logger.info('Waiting for %s seconds...', str(sec))
         time.sleep(sec)
 
+    def proof_connect(self, output):
+        """Proof if connection output is not containing errors"""
```
ajinabraham: """Check if connect failed"""

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
+        if b'unable to connect' in output or b'failed to connect' in output:
+            logger.error('%s', output.decode('utf-8').replace('\n', ''))
+            return False
+        else:
```
ajinabraham: `else` is not needed. `return True` should do.

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
         self.wait(2)
         logger.info('Connecting to Android %s', self.identifier)
-        out = subprocess.check_output([get_adb(), 'connect', self.identifier])
-        self.wait(2)
-        if b'unable to connect' in out or b'failed to connect' in out:
-            logger.error('%s', out.decode('utf-8').replace('\n', ''))
+        if not run_subprocess_verify_output([get_adb(),
+                                            'connect',
+                                             self.identifier]):
             return False
```
ajinabraham: `return self. run_subprocess_verify_output..`

DynamicAnalyzer/views/android/environment.py
```diff
@@ -40,41 +40,49 @@ def wait(self, sec):
...
-                                           self.identifier])
-            self.wait(2)
-            if b'unable to connect' in out or b'failed to connect' in out:
-                logger.error('%s', out.decode('utf-8').replace('\n', ''))
-                return False
-            # mount system
-            logger.info('Remounting /system')
-            self.adb_command(['mount', '-o',
-                              'rw,remount', '/system'], True)
+        # TODO: proof if not already connected as admin
```
ajinabraham: Remove comment if not related or update accordingly.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: Done

DynamicAnalyzer/views/android/environment.py
toxicphreAK: you are right, I was to quick. sorry for the confusion!

DynamicAnalyzer/views/android/environment.py
toxicphreAK: yes

DynamicAnalyzer/views/android/environment.py
toxicphreAK: yes

DynamicAnalyzer/views/android/environment.py
toxicphreAK: good idea, this makes it even smaller

DynamicAnalyzer/views/android/environment.py
toxicphreAK: changed it

DynamicAnalyzer/views/android/environment.py
toxicphreAK: You are right.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: Then the code will stop after that, isn't it? So I just want to only return False if it is False, otherwise continue and return True at the end.

DynamicAnalyzer/views/android/environment.py
toxicphreAK: removed

DynamicAnalyzer/views/android/environment.py
toxicphreAK: removed

toxicphreAK:
Sorry for the bad last changes. Hopefully I will get better time after time, as I collected some development ideas in my fork: https://github.com/toxicphreAK/Mobile-Security-Framework-MobSF/issues trying to support your nice work!
If you have any tips in sense of readings, etc. to get better, you are very welcome!

DynamicAnalyzer/views/android/environment.py
ajinabraham: Yes you are right.

ajinabraham:
Good to merge

PR #1234, opened 2020-01-22
Remove scan type from reports

Nicoolai:
### Describe the Pull Request

```
Currently for zipped source files, the upload api doesn't return what type of zip it is. 
This needs to be explicitly set or scan should be able to happen without specifying the type of zip, ie andzip or ioszip .

I have changed the report calls to try all static_db's in search for the report. It then uses whatever returns a result, making it able to automatically figure out the source type.
```

### Checklist for PR

- [ ] Make sure build is passing on your PR. [Build Status](https://travis-ci.com/MobSF/Mobile-Security-Framework-MobSF/pull_requests)

```
I have also updated tests and API docs.
```


StaticAnalyzer/tests.py
```diff
@@ -52,18 +52,18 @@ def static_analysis_test():
         logger.info('Running PDF Generation Test')
         if platform.system() in ['Darwin', 'Linux']:
             pdfs = [
-                '/PDF/?md5=3a552566097a8de588b8184b059b0158&type=APK',
```
Nicoolai: I am personally a bit uncertain about these changes.  
Not entirely sure if this "PDF Generation Test" is something that should be changed.  
So please, add any feedback regarding this.

ajinabraham: Looks good to me. We want to run PDF generation tests for all kind of source formats we support.

ajinabraham:
Code looks good to me. Testing this out. Thanks for the PR

ajinabraham:
Testing went good. great work!

Issue #1235, opened 2020-01-24
Further clarification on licensing terms


oxr463:
Is the license GPL-3.0-only, or GPL-3.0-or-later?

See: https://www.gnu.org/licenses/identify-licenses-clearly.html

superpoussin22:
Clarification Done --- clsoing

PR #1236, opened 2020-01-24
Use SPDX identifier for license

oxr463:
Bug: https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/1235
See: https://spdx.org/licenses/GPL-3.0-only.html


ajinabraham:
`GPL--3.0--only` sounds good to me.

PR #1288, opened 2020-03-02
Prepare MobSF to run Swift with existing rules.

karolpiateknet:
- [x] Prepare MobSF to run Swift with existing rules.
- [x] Create separate rules - common, objc, swift.
- [x] Add DVIA-V2 as Swift test.
- [x] Add DVIA-V2 hash to tests.

### Checklist for PR


StaticAnalyzer/views/ios/code_analysis.py
```diff
@@ -70,7 +76,6 @@ def ios_source_analysis(src):
             'urlnfile': url_n_file,
             'domains': domains,
             'emailnfile': email_n_file,
-            'source_type': source_type,
```
ajinabraham: We need a logic to detect the kind of source here.
1. Swift
2. Objective-C
3. Swift, Objective-C

StaticAnalyzer/views/ios/common_rules.py
```diff
@@ -0,0 +1,137 @@
...
+        'input_case': 'lower',
+        'cvss': 7.4,
+        'cwe': 'CWE-312',
+        'owasp': 'M9: Reverse Engineering',
+    },
+    {
+        'desc': ('App uses SQLite Database. '
+                 'Sensitive Information should be encrypted.'),
+        'type': 'string',
+        'string1': 'sqlite3_exec',
```
ajinabraham: Is this applicable to Swift?

StaticAnalyzer/views/ios/common_rules.py
```diff
@@ -0,0 +1,137 @@
...
+        'input_case': 'exact',
+        'cvss': 0,
+        'cwe': '',
+        'owasp': '',
+    },
+    {
+        'desc': ('User input in "loadHTMLString" '
+                 'will result in JavaScript Injection.'),
+        'type': 'string',
+        'string1': 'loadHTMLString',
```
ajinabraham: Is this applicable to Swift?

StaticAnalyzer/views/ios/swift_rules.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+
+from StaticAnalyzer.views.ios import common_rules
+
+SWIFT_RULES = common_rules.COMMON_RULES + [
+    {
+        'desc': ('The App logs information. '
+                 'Sensitive information should never be logged.'),
+        'type': 'regex',
+        'regex1': r'print|NSLog|os_log|OSLog|os_signpost',
```
ajinabraham: Prone to lot of false positives. This needs to be fine tuned to something like `print(*)` or `NSLog(` etc.

StaticAnalyzer/views/ios/swift_rules.py
```diff
@@ -0,0 +1,114 @@
...
+        'input_case': 'exact',
+        'cvss': 7.5,
+        'cwe': 'CWE-532',
+        'owasp': '',
+    },
+    {
+        'desc': 'SHA1 is a weak hash known to have hash collisions.',
+        'type': 'regex',
+        'regex1': r'(?i)SHA1\(',
+        'regex2': r'CC_SHA1',
```
ajinabraham: make the regex to check for `CC_SHA1(` to reduce false positives

StaticAnalyzer/views/ios/swift_rules.py
```diff
@@ -0,0 +1,114 @@
...
+        'input_case': 'exact',
+        'cvss': 5.9,
+        'cwe': 'CWE-327',
+        'owasp': 'M5: Insufficient Cryptography',
+    },
+    {
+        'desc': 'MD2 is a weak hash known to have hash collisions.',
+        'type': 'regex',
+        'regex1': r'(?i)MD2\(',
+        'regex2': r'CC_MD2',
```
ajinabraham: make the regex to check for `CC_MD2(` to reduce false positives

StaticAnalyzer/views/ios/swift_rules.py
```diff
@@ -0,0 +1,114 @@
...
+        'input_case': 'exact',
+        'cvss': 5.9,
+        'cwe': 'CWE-327',
+        'owasp': 'M5: Insufficient Cryptography',
+    },
+    {
+        'desc': 'MD4 is a weak hash known to have hash collisions.',
+        'type': 'regex',
+        'regex1': r'(?i)MD4\(',
+        'regex2': r'CC_MD4',
```
ajinabraham: make the regex to check for `CC_MD4(` to reduce false positives

StaticAnalyzer/views/ios/swift_rules.py
```diff
@@ -0,0 +1,114 @@
...
+        'input_case': 'exact',
+        'cvss': 5.9,
+        'cwe': 'CWE-327',
+        'owasp': 'M5: Insufficient Cryptography',
+    },
+    {
+        'desc': 'MD5 is a weak hash known to have hash collisions.',
+        'type': 'regex',
+        'regex1': r'(?i)MD5\(',
+        'regex2': r'CC_MD5',
```
ajinabraham: make the regex to check for `CC_MD5(` to reduce false positives

StaticAnalyzer/views/ios/swift_rules.py
```diff
@@ -0,0 +1,114 @@
...
+        'input_case': 'exact',
+        'cvss': 5.9,
+        'cwe': 'CWE-327',
+        'owasp': 'M5: Insufficient Cryptography',
+    },
+    {
+        'desc': 'MD6 is a weak hash known to have hash collisions.',
+        'type': 'regex',
+        'regex1': r'(?i)MD6\(',
+        'regex2': r'CC_MD6',
```
ajinabraham: make the regex to check for `CC_MD6(` to reduce false positives

StaticAnalyzer/views/ios/common_rules.py
karolpiateknet: Yes.
Reference: https://github.com/OWASP/owasp-mstg/blob/master/Document/0x06h-Testing-Platform-Interaction.md#testing-how-webviews-are-loaded

StaticAnalyzer/views/ios/code_analysis.py
karolpiateknet: Done

StaticAnalyzer/views/ios/common_rules.py
karolpiateknet: Yes, but I see that sqlite3 may be used with different function than exec.
Maybe we should detect patter like `sqlite3_****(` with regex `sqlite3[_][\S]*\(`.
So I could leave the `sqlite3_exec` in objc and add this regex in Swift.

StaticAnalyzer/views/ios/swift_rules.py
karolpiateknet: Changed to: `(print|NSLog|os_log|OSLog|os_signpost)\(.*\)` It will match print(.*) and others.

StaticAnalyzer/views/ios/code_analysis.py
```diff
@@ -2,62 +2,85 @@
...
+                urls, urls_nf, emails_nf = url_n_email_extract(
+                    dat, relative_src_path)
+                url_list.extend(urls)
+                url_n_file.extend(urls_nf)
+                email_n_file.extend(emails_nf)
+
+        if len(source_types) > 1:
+            source_type = _SourceType.swift_and_objc.value
+        else:
+            source_type = source_types.pop().value
```
ajinabraham: What if the zip package had no swift or objective c file (An edge case)?
 `source_types` will be an empty set. We need MobSF to handle exception gracefully and proceed with the rest of the checks even if one the process failed.

StaticAnalyzer/views/ios/common_rules.py
ajinabraham: We only need to consider the `sqlite3_**` functions that execute raw SQL query. If you know of other functions with the prefix, we should consider them. My question was more like is the function `sqlite3_exec` having same syntax/function name is Swift as well?

karolpiateknet: Yes, sqlite3_exec is applicable in Swift, It has the same syntax.
To run sqlite statement We have to execute `sqlite3_finalize` function, I will add it with next part of the rules.

PR #1306, opened 2020-03-06
Create Match enum

karolpiateknet:
### Create Match enum

```
Match is an Enum used in static code analysis to run proper rule detection.
Such approach will show all available options.
 It will eliminate risk of wrong string, since it won't run with wrong enum case.
In the future most of the params from rules could be represented as enums.
```

### Checklist for PR


ajinabraham:
I like this 💯 !

PR #1322, opened 2020-03-11
iOS: Swift TLS rules

mtbrzeski:
### Describe the Pull Request

```
Add CWE-310 standard - Cryptographic Issues
Add rules for TLS/DTLS versions
```

### Checklist for PR

```
@ajinabraham please verify if CWE-310 standard applies properly to the new rules, which check for proper TLS version. Reference: https://www.cvedetails.com/cve/CVE-2015-4000/

I've used **high** for TLS 1.0, TLS 1.1 and DTLS 1.0, along with CVSS score and CWE category
**warning** is used for TLS 1.2, without CVSS score nor CWE category
**info** is triggered if a deprecated property is used to set SSL version
```


StaticAnalyzer/views/ios/rules/swift_rules.py
```diff
@@ -375,4 +375,74 @@
...
+        'type': MatchType.string,
+        'string1': '.tlsMinimumSupportedProtocol =',
+        'level': Level.info,
+        'match': Match.single_string,
+        'input_case': InputCase.exact,
+        'cvss': 0.0,
+        'cwe': '',
+        'owasp': OWASP['m3'],
+        'owasp-mstg': OWASP_MSTG['network-2'],
+    },
```
karolpiateknet: I'm just wondering if we could merge all these rules into one rule like:
```
{
        'desc': ('App may use old network encryption',
                     ' network encryption should always be in the newest version.'),
        'type': MatchType.string,
        'string1': '.tlsMinimumSupportedProtocol =',
        'string2': '.TLSMinimumSupportedProtocolVersion =',
        'level': Level.warning,
        'match': Match.string_or,
        'input_case': InputCase.exact,
        'cvss': 4.3,
        'cwe': CWE['CWE-310'],
        'owasp': OWASP['m3'],
        'owasp-mstg': OWASP_MSTG['network-2'],
},
```
What do you think?
@mtbrzeski  @ajinabraham

mtbrzeski: Would be nice to merge, but there are a couple of issues.

- **warning** is always used, while in the original code there can be either info, warning, or high depending on the TLS/DTLS version. Same with CVSS score
- you can have a false-positive here (which assigns CVSS) if somebody assigns TLS 1.3 manually

Maybe we could merge the rule for TLS 1.0 and TLS 1.1 as one? And keep the rest for precision. Also depends on the input from @ajinabraham regarding _levels_ and assigning _cvss_ to each case.

StaticAnalyzer/views/ios/rules/swift_rules.py
```diff
@@ -375,4 +375,74 @@
...
+    {
+        'desc': ('Use of deprecated property tlsMinimumSupportedProtocol. '
+                 'To avoid potential security risks, use '
+                 'tlsMinimumSupportedProtocolVersion'),
+        'type': MatchType.string,
+        'string1': '.tlsMinimumSupportedProtocol =',
+        'level': Level.info,
```
ajinabraham: Can this be changed to warning?

StaticAnalyzer/views/ios/rules/swift_rules.py
ajinabraham: @mtbrzeski I like the idea of merging TLS1.0 and 1.1. Since they both share the level high and TLS 1.2 is of level medium.
Can we change the match to
`string_and_or` for TLS1.0,1.1
and
`string_and` for TLS1.2 and DTLS 1.2

Let's avoid ambiguity from `=` and whitespace
Something like
`string1`: `TLSMinimumSupportedProtocolVersion`
`string2`: `tls_protocol_version_t.TLSv12`

StaticAnalyzer/views/standards.py
```diff
@@ -129,6 +129,7 @@
     'CWE-250': 'CWE-250 - Execution with Unnecessary Privileges',
     'CWE-276': 'CWE-276 - Incorrect Default Permissions',
     'CWE-295': 'CWE-295 - Improper Certificate Validation',
+    'CWE-310': 'CWE-310 - Cryptographic Issues',
```
ajinabraham: CWE 757 is more suitable.

https://cwe.mitre.org/data/definitions/757.html

PR #1426, opened 2020-06-09
Add Bouncy Castle and Jackson libraries to whitelisted list

peja:
The Bouncy Castle and Jackson JSON library are widely used, and they should be included in SKIP_CLASSES list of common third party classes that should be skipped during static analysis.


MobSF/settings.py
```diff
@@ -251,11 +251,13 @@
         r'org[\\\/]{1}chromium[\\\/]{1}',
         r'com[\\\/]{1}facebook[\\\/]{1}',
         r'org[\\\/]{1}spongycastle[\\\/]{1}',
+        r'org[\\\/]{1}bouncycastle[\\\/]{1}',
         r'com[\\\/]{1}amazon[\\\/]{1}identity[\\\/]{1}',
         r'io[\\\/]{1}fabric[\\\/]{1}sdk[\\\/]{1}',
         r'com[\\\/]{1}instabug[\\\/]{1}',
         r'io[\\\/]{1}fabric[\\\/]{1}sdk[\\\/]{1}',
         r'com[\\\/]{1}crashlytics[\\\/]{1}android[\\\/]{1}',
+        r'com[\\\/]{1}fasterxml[\\\/]{1}jackson[\\\/]{1}',
```
ajinabraham, line 260: We can't include jackson since we check for a vulnerability in the same.

https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/StaticAnalyzer/views/android/rules/android_rules.py#L617

peja: I don't think those two are related: We should definitely check if someone is using the library in an unsafe way, but we shouldn't scan the library itself.

ajinabraham: Oh yes you are right. I got caught up seeing Jackson as I remember we had a rule for the same. 🤦

peja: It's okay :smiley: Take care!

PR #1561, opened 2020-10-11
Fixed permission levels and descriptions according to Google documentation

pavelkrcma:
Description and Protection level for many permissions currently does not correspond to up-to-date Google documentation available here: [Manifest.permission](https://developer.android.com/reference/android/Manifest.permission). This PR brings up-to-date texts and protection levels for all permissions.


ajinabraham (approved):
Thanks for the PR @pavelkrcma
LGTM 👍

PR #1601, opened 2020-11-24
website display error in debug mode

XmTX1994:
```
When running in debug mode, page cannot display properly.
```

### DIFF
--- a/MobSF/settings.py
+++ b/MobSF/settings.py
@@ -204,7 +204,7 @@ TEMPLATES = [
 MEDIA_ROOT = os.path.join(BASE_DIR, 'uploads')
 MEDIA_URL = '/uploads/'
 STATIC_URL = '/static/'
-STATIC_ROOT = os.path.join(BASE_DIR, 'static')
+STATICFILES_DIRS = (os.path.join(BASE_DIR, 'static'), )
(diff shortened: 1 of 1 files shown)


ajinabraham:
I am sorry, could you please briefly explain about this debug mode issue that needs this fix?

MobSF by default runs on debug mode and haven't noticed any issues.
https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/master/MobSF/settings.py#L152

XmTX1994:
In settings.py file, when I set DEBUG = True, page like this:
image

So, I think following code should be added
if DEBUG:
    STATICFILES_DIRS = (os.path.join(BASE_DIR, 'static'), )
else:
    STATIC_ROOT = os.path.join(BASE_DIR, 'static')

ajinabraham:
When I am running as per the docs. https://mobsf.github.io/docs/#/running
I do not see any issue with `STATIC_ROOT = os.path.join(BASE_DIR, 'static')`

XmTX1994:
Oh, I used "python manage.py runserver" to run mobsf. When I am using setup.bat and run.bat, it works well.

ajinabraham:
Closing this.

PR #1723, opened 2021-04-19
Update manifest_analysis.py

firmianay:
just delete `itemname = 'NIL'` branch and left shift the following code to make it more clear.


ajinabraham:
Thanks for the PR @firmianay I will review this soon.

PR #1740, opened 2021-05-28
Include browsable activities in comparer

santoru:
### Describe the Pull Request
Comparer is really helpful to analyse differences between applications but it lack some important features.
One of these is the ability to detect differences between browsable activities (deeplinks/universal links) in Android, which is a big attack surface and it can be very interesting to monitor and to detect additions.
Since browsable activities are more complex objects that others (i.e. activities) we need to perform comparison not only on the first level (which browsable activities are in common and which not) but also in deep: from the browsable activities in common, which schemes, hosts, paths are in common and which not?

This PR include a new diffing function for common browsable activities that explore each activity in both apps to detect differences for a set of keys (scheme/hosts etc..) and include them in the context. 
Additionally, browsable activities that are not in common are added to the proper `only_first` or `only_second` component.
The template is responsible to display these information: for each browsable activity a table is shown with a row header indicating the browsable activity, and the information are shown like for APKiD information. For common browsable activities the table has 3 rows to show differences between the components of the browsable activity, as shown on this example:

br_act

### Checklist for PR

### DIFF
--- a/mobsf/StaticAnalyzer/views/comparer.py
+++ b/mobsf/StaticAnalyzer/views/comparer.py
@@ -64,6 +64,7 @@ def diff_apkid(context: dict) -> None:
             x for x in tmp_flat_results['second_app'][key] if x not in
             tmp_flat_results['first_app'][key]]
 
+
 def diff_browsable_activites(context, first_app, second_app):
     # Browsable activites are 'complex' objects that contains additional
     # fields that should be compared. The generic diffing performed above
@@ -71,8 +72,8 @@ def diff_browsable_activites(context, first_app, second_app):
     # one app. For the activites present in one app we can just pretty-print
     # the details but for common browsable activities we need to perform the
     # diffing also on sub-keys, like hosts and schemes.
-    section = "browsable_activities"
-    common_activity = "common_browsable_activities"
+    section = 'browsable_activities'
+    common_activity = 'common_browsable_activities'
     browsable_activities_keys = ['schemes', 'mime_types', 'hosts', 'ports',
                                  'paths', 'path_prefixs', 'path_patterns']
 
--- a/mobsf/StaticAnalyzer/views/comparer.py
+++ b/mobsf/StaticAnalyzer/views/comparer.py
@@ -64,6 +64,38 @@ def diff_apkid(context: dict) -> None:
             x for x in tmp_flat_results['second_app'][key] if x not in
             tmp_flat_results['first_app'][key]]
 
+def diff_browsable_activites(context, first_app, second_app):
+    # Browsable activites are 'complex' objects that contains additional
+    # fields that should be compared. The generic diffing performed above
+    # only list which browsable activities are in common or present only in
+    # one app. For the activites present in one app we can just pretty-print
+    # the details but for common browsable activities we need to perform the
+    # diffing also on sub-keys, like hosts and schemes.
+    section = "browsable_activities"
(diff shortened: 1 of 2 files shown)


ajinabraham:
Hi @santoru ,
Thanks for the PR. ❤️ 
I will take a look at this soon.

ajinabraham (commented):
I tested this out and it looks like it's not displayed properly

Screen Shot 2021-06-04 at 6 23 07 PM

It would be nice to have a table with 3 columns 
`Only in app a | only in app b | common`

You could just re-use the table format.

https://github.com/MobSF/Mobile-Security-Framework-MobSF/blob/36b089ad1e643ca5d4913941b7d7d8be1b5df3a4/mobsf/templates/static_analysis/android_binary_analysis.html#L807-L866

santoru:
Hi @ajinabraham. Thank you for the review. Let me check it and update the PR, thank you

santoru:
Hi @ajinabraham,
The issue I saw with the 3 columns design while making the PR was that it contains multiple white cells when the compared apps have not common activities.
Let's say I have two apps A and B with 3 exported activities, 2 in commons and 1 only present in app A:
The 3 columns design will have two empty cells for the table of the not shared activity, filling the information only for the app A. The design I choose, taken from the ApkID one, will instead only have 1 row for the app that has the activity, while the other one will not be displayed. 
Additionally, for shared exported activities, you will have more comparison material here:
You will have the 3 rows: 
```
Common
-
Only in app A
-
Only in app B
```
while the columns keep the consistency with the other tables and display deeper details like hosts, schemes, paths.

What do you think? Does this make sense to you?
Or do you meant something else?
Thank you.

ajinabraham:
Since this is app based diffing. I prefer them to follow the similar patterns to the rest of the findings.
Also I am not clear upon what you are diffing things here. Is it based on activity name or the host/schemas/others?
Nevertheless, it's less confusing to follow the existing pattern. You don't have to create additional columns for say mime, paths etc. that might not be present always, but rather show what's present in just a column (assuming the diffing is based on activity name)

This is a quick mock I had in mind

Screen Shot 2021-06-11 at 10 48 00 AM-2

Also what's a valid use case? Is this to compare 2 versions of the same app or two different apps?
If you would like to talk on this further and get it clarified on both side, we could quickly setup a call. I am available from 3pm - 6pm PT throughout the week. <ajin25 AT gmail DOT com>

santoru:
Thank you @ajinabraham for the details, I overlooked the template you linked before and I didn't notice it fit this use case very good. I will change the design of the table to reflect this template, which is indeed more clean and more coherent with the existing design.

About the use case, I'm currently using extensively the comparer for these main (similar) reasons:
- As a security researcher / bug bounty hunter I would like to detect new deeplinks when a new app version is released, as these are a big attack surface (so I do the comparison between same app, different version).
- As a security engineer I want to review internal versions of the application before the release, in a way that added components are tested properly :) 

Thank you again for your help

ajinabraham:
@santoru got sidetracked with work. I will test this again soon.

george79d (approved):

george79d (dismissed):

ajinabraham (approved):
Tested this. Looks good! 👍

ajinabraham (approved):

ajinabraham:
Thank you for the PR and your patience. This is merged to master. 🎊

santoru:
Thank you for the review and your help!

PR #1812, opened 2021-09-23
Run mobsf in dockerfile as unprivileged user and fix a few minor bugs.

Alexeyan:
This pull request changes the Dockerfile to run the service without root privileges.
This improves security, as a docker container is [not a security boundary](https://docs.docker.com/engine/security/) and this project is used to analyse potentially malicious samples. 

It also fixes a bug in that ubuntu doesn't allow pip to update itself when it's installed by apt. ([Reference](https://stackoverflow.com/questions/53807511/pip-cannot-uninstall-package-it-is-a-distutils-installed-project/53807588)) 
It also bumps the version of dependency click, as the docker build warned about incompatible versions before. 

Best regards


ajinabraham (approved):
LGTM  :+1

PR #1845, opened 2021-11-18
it's better not to use `apt` in scripts

aminvakil:
### Describe the Pull Request
Taken from man apt:
```
apt provides a high-level commandline interface for the package management system. It is intended as an end user interface and enables some options better suited for interactive usage by default compared to more specialized APT tools like apt-get(8) and apt-cache(8).
```

### Checklist for PR

- [ ] Run MobSF unit tests and lint `tox -e lint,test`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,7 +22,7 @@ ENV DEBIAN_FRONTEND="noninteractive" \
 ENV PATH="$JAVA_HOME/bin:$PATH"
 
 # See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run
-RUN apt update -y && apt install -y  --no-install-recommends \
+RUN apt-get update -y && apt-get install -y  --no-install-recommends \
     build-essential \
     locales \
     sqlite3 \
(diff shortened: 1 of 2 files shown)


ajinabraham:
`apt should not be used in scripts` is there a documentation that states `apt` should not be used in cli / dockerfile?

aminvakil:
> `apt should not be used in scripts` is there a documentation that states `apt` should not be used in cli / dockerfile?

As mentioned in PR description, here is also a link:
https://manpages.debian.org/bullseye/apt/apt.8.en.html

apt provides a high-level commandline interface for the package management system. It is intended as an end user interface and enables some options better suited for interactive usage by default compared to more specialized APT tools like apt-get(8)and apt-cache(8).

ajinabraham:
Yes, but there is nothing that says you should not use `apt` in scripts.

aminvakil:
> Yes, but there is nothing that says you should not use `apt` in scripts.

I changed title of my PR.

ajinabraham:
Unless there are issues with using `apt`, I am reluctant towards this change as this does not bring any improvement on what's already there.

aminvakil:
`apt` does not need to be backward-compatible like `apt-get` should be, I'm not saying that maybe someday `apt install package -y` will not work, but it feels better to use something that is intended to backward compatible.

ajinabraham:
I am closing this PR as the change does not particularly fix anything that's broken now or documented to be broken in the near future. We can always revisit once it's officially recommend to use `apt-get` instead of `apt` for CLI.

PR #1919, opened 2022-03-04
Improved plist parser

pavelkrcma:
### Describe the Pull Request

This PR implements two major improvements in plist parser:

* parser for SKAdNetworkItems introduced in iOS 11.3 (https://developer.apple.com/documentation/bundleresources/information_property_list/skadnetworkitems) including list of known ad network identifiers
* machine-readable ATS structure. Currently mobSF provides only result of analysis of ATS but not raw ATS structure which is useful for further processing

### Checklist for PR

- [X] Run MobSF unit tests and lint `tox -e lint,test`
- [X] Make sure tests are passing on your PR [MobSF tests](https://github.com/MobSF/Mobile-Security-Framework-MobSF/actions)


pavelkrcma:
@ajinabraham please check

pavelkrcma:
@ajinabraham could you please review this functionality?

mobsf/StaticAnalyzer/views/ios/adnetworkidentifiers.py
```diff
@@ -0,0 +1,56 @@
+known_adnetworks = {
```
ajinabraham, line 1: What is the source of this list?

mobsf/StaticAnalyzer/views/ios/db_interaction.py
```diff
@@ -38,6 +38,8 @@ def get_context_from_db_entry(db_entry):
             'binary_info': python_dict(db_entry[0].BINARY_INFO),
             'permissions': python_dict(db_entry[0].PERMISSIONS),
             'ats_analysis': python_list(db_entry[0].ATS_ANALYSIS),
```
ajinabraham, line 40: Can we reuse `ats_analysis` for storing `transport_security_info` and `adnetworkidentifiers`. I would prefer not to create a new model field unless it's absolutely necessary.

something like 

```
ats_analysis : { 

  'ats': list(),
  'transportsecurityinfo': list(),
  'adnetworkidentifiers': dict{},
  }
  ```

ajinabraham (commented):
Thanks @pavelkrcma for the PR. This looks interesting and useful.

Few questions and minor suggestions.
Also, how can we integrate these results to UI?

mobsf/StaticAnalyzer/views/ios/adnetworkidentifiers.py
pavelkrcma: There are organisations collecting these identifiers:

https://www.skanids.com/
https://github.com/skadnetwork/ad-network-ids

pavelkrcma:
Regarding integration to UI - I'm not sure. I consider SKAdNetworkItems parser more like preparation for more advanced parsing in future but currently I don't have any clue what else we can get from them. This information would be interesting if we identify that e.g. one of those ad networks serves more malicious or explicit content than other. But currently I don't have such data.

ajinabraham:
Unless we can generate some useful security metrics out of the data collected, I don't think we will be able to merge this. All of these information are already available in `INFO_PLIST` column for further parsing/processing in future.

The ad network identifier list in the PR is a subset of https://www.skanids.com/ and I imagine this list to grow over time. If we are concerned about a malicious ad network identifier, the list should only include the malicious ones as and when they are discovered. Going the other way is a maintenance nightmare for an open source project.

We already does security analysis on ATS data, keeping a parsed copy is redundant unless it is connected to a security metrics.

pavelkrcma:
OK, understand. So we create our own clone with this change.

PR #1954, opened 2022-05-02
Fixed python version checking and OpenSSL broken link.

TSlayman:
Added python 3.10+ checking and updated the link for OpenSSL.exe.

The python-checking code was refusing to resume the setup if the version is above 3.10+.
Also updated the link to OpenSSL.exe application.

### DIFF
--- a/setup.bat
+++ b/setup.bat
@@ -13,7 +13,7 @@ where python >nul 2>&1 && (
   ) else (
     exit /b
   )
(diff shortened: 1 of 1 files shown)


superpoussin22 (commented):
you must also change the error message

TSlayman:
> you must also change the error message 

That's right, totally forgot about it.

PR #1905, opened 2022-02-16
Add the checking of the parent element of the permission-related elements to manifest analysis

Han0nly:
### Describe the Pull Request

```
Hi, The parent element of <permission>,<uses-permission>, <uses-permission-sdk-23> elements can only be <manifest>, or these statements would not take effect. According to an empirical study of a paper, this kind of bug is very common. So in this pull request I add the checking of the parent node of these elements to the manifest analysis part.
```

### Checklist for PR

- [*] Run MobSF unit tests and lint `tox -e lint,test`
- [*] Tested Working on Linux, Mac, Windows, and Docker
- [*] Add unit test for any new Web API (Refer: `StaticAnalyzer/tests.py`)
- [*] Make sure tests are passing on your PR [MobSF tests](https://github.com/MobSF/Mobile-Security-Framework-MobSF/actions)

### Additional Comments (if any)

```
In this pull request, I only add the checking of the permission elements, but I'll continuously add all the other manifest structure checking functionalities later.
```

PR #1984, opened 2022-06-24
Revert "Add the checking of the parent element of the permission-related elements to manifest analysis"

ajinabraham:
Reverts MobSF/Mobile-Security-Framework-MobSF#1905

Causing a lot of false positives


ajinabraham:
HOTFIX: Revert #1905

PR #2172, opened 2023-04-25
Dynamic Analysis from Docker Container using remote ADB debugging over TCP/IP

JohLud:
Dynamic Analysis is not supported from a docker container. This is because the environment in DynamicAnalyzer/views/android/environment.py is configured to only use serial device identifier (`adb -s <id>`). But it is also possible to allow ADB to debug over TCP/IP and thus communicate with an emulator instance and runs either on the host system or even also in a container. This is done by using (`adb -H <ip address>`).

For this, the following changes have been made:
* add a new environment variable `ADB_REMOTE_HOST` (analogous to `DEVICE_IDENTIFIER`)
* in environment.py check if this variable is set, and if this is the case:
* run adb commands over TCP/IP with a new command line: `adb -H ADB_REMOTE_HOST <commands>`

This new variable should be the IP Address of the host system. The host system should run an ADB daemon and an emulator instance.
Per default the ADB Daemon listens on localhost, therefore forwarding from the external network interface (eth0) to the loopback interface (lo0) should be configured.
This must be done for the following ports: 5037, 5554, 5555

If `ADB_REMOTE_HOST` environment variable is set (either in config.py or with `docker -r ADB_REMOTE_HOST`), by starting the dynamic analyzer, mobsf should automatically detect the running emulator instance and MobSFying and debugging should be possible.

Tested with Android Emulator and AVD Nexus_One_API_28 on Linux.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -13,6 +13,7 @@ LABEL \
 # Environment vars
 ENV DEBIAN_FRONTEND="noninteractive" \
     ANALYZER_IDENTIFIER="" \
+    ADB_REMOTE_HOST="" \
     JDK_FILE="openjdk-16.0.1_linux-x64_bin.tar.gz" \
     JDK_FILE_ARM="openjdk-16.0.1_linux-aarch64_bin.tar.gz" \
     WKH_FILE="wkhtmltox_0.12.6-1.focal_amd64.deb" \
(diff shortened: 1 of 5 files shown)


ajinabraham:
Hi @JohLud Thanks for the PR. I will take a look and get back.

ajinabraham:
We are using a different set of instructions to talk with AVD from docker: https://mobsf.github.io/docs/#/dynamic_analyzer?id=android-studio-emulator

PR #2176, opened 2023-05-15
Fix missing exported components

Abb4d0n:
This changes fix an issue in the Android manifest analysis where exported components which have no permission were not listed in the results because of a wrong template description key. Also added a warning if this happens again.


ajinabraham (approved):
Late response, but this looks good.

Issue #2286, opened 2023-11-15
oscrypto / openSSL regex bug when running setup.sh


SpookyB00k:
## ENVIRONMENT

```
OS and Version: Kali Linux Rolling Release 6.5.0-kali3-amd64
Python Version: 3.11.6
MobSF Version: 3.7.9 Beta
```

## EXPLANATION OF THE ISSUE

When trying to install MobSF with the provided setup.sh, you may run into the issue that the install fails because it is unable to detect the version of libcrypto.
![image](https://github.com/MobSF/Mobile-Security-Framework-MobSF/assets/37688484/adf1704c-3261-4c27-8be2-bcddf9b02a59)
When working with versions that are higher or equal than 3.0.10 of OpenSSL, the Regex used to detect does not work.
Currently it is 
```
\b(\d\.\d\.\d[a-z]*)\b
```

Personally I monkeypatched /home/Sp00kyB00k/.cache/pypoetry/virtualenvs/mobsf-vfi9xHNf-py3.11/lib/python3.11/site-packages/oscrypto/_openssl/_libcrypto_cffi.py to which is far from perfect but good enough to get the install rolling.
```
\b(\d\.\d\.\d+[a-z]*)\b
```
Adding the extra + with the last digit should fix it. 

## STEPS TO REPRODUCE THE ISSUE

Install on Kali Linux by cloning the repository and running the setup.sh

danilabs:
Awesome @SpookyB00k you save me a lot of time :clap: !
Thank you!

ajinabraham:
We are removing oscrypto dependency from MobSFv3.8 onwards. That should address this.

baptx:
Thanks, the solution worked.
For information, I also had an error "externally-managed-environment" that I fixed by modifying setup.sh to add `--break-system-packages` at the end of each pip line.
More details here: https://stackoverflow.com/questions/75608323/how-do-i-solve-error-externally-managed-environment-every-time-i-use-pip-3/75722775#75722775

PR #2288, opened 2023-11-16
Update pyproject.toml to fix issue #2286

SpookyB00k:
### Describe the Pull Request

Fixed a bug in oscrypto where the regex would fail to recognize the right version of SSL due to insufficient check on multiple digits.  However the owner of the repo does not plan to make a new release soon due to time constraints.  See -> https://github.com/wbond/oscrypto/issues/78 

This can be fixed by using a specific version from git pointing to the right commit.  Hence this change in the pyproject.toml file.

### Checklist for PR

Tested on the Kali System mentioned in Issue #2286.
If this is not good enough for this change, I need to invest more time.
However this is an upstream fix in the lib itself and only changes the dependency to a specific commit.

https://github.com/MobSF/Mobile-Security-Framework-MobSF/issues/2286

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,7 +38,7 @@ gunicorn = {version = ">=20.0.4", markers = "sys_platform != 'win32'"}
 psutil = ">=5.8.0"
 shelljob = ">=0.6.2"
 asn1crypto = ">=1.4.0"
-oscrypto = ">=1.2.1"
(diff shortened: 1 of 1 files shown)


ajinabraham:
We will be [removing](https://github.com/MobSF/Mobile-Security-Framework-MobSF/pull/2194/commits/33750714bb4b27e6be6cca9bf00bdda108ad90cd) the dependency on `oscrypto` from MobSF v4 onwards.

PR #2351, opened 2024-03-04
Optimize rendering of big lists

JPSxzy8:
### Describe the Pull Request

Most browsers doesn't like rendering thousands/millions of lines if not needed, so wrap (potentially) big lists in `<details>` and auto-open only if they are small (currently below 100 items)

This makes the "static report" page much more responsive for big apps:

image

### Checklist for PR

- [x] Run MobSF unit tests and lint `tox -e lint,test`
- [x] Tested Working on Linux, ~Mac, Windows, and Docker~ (this PR only adds cosmetic changes to HTML)
- [x] Add unit test for any new Web API (Refer: `StaticAnalyzer/tests.py`) (this PR only adds cosmetic changes to HTML, so no new tests)
- [x] Make sure tests are passing on your PR [MobSF tests](https://github.com/MobSF/Mobile-Security-Framework-MobSF/actions)

### DIFF
--- a/mobsf/templates/static_analysis/android_binary_analysis.html
+++ b/mobsf/templates/static_analysis/android_binary_analysis.html
@@ -2118,7 +2118,7 @@
               <strong><i class="fas fa-key"></i> POSSIBLE HARDCODED SECRETS</strong>
               </p>
                 <div class="list-group">
-                  {% include 'base/list.html' with list=secrets type="secrets" %}
+                  {% include 'base/list.html' with list=secrets type="secrets" limit=100 %}
                </div>
              </div>
            </div>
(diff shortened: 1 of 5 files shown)


ajinabraham:
Nice! I will test this and get back.

mobsf/templates/static_analysis/android_binary_analysis.html
```diff
@@ -2145,28 +2141,13 @@ <h5 class="description-header">{{ code_analysis.summary.suppressed }}</h5>
...
-                      {% endif %}
-                    {% endfor %}                  
-                  </p>
+                  {% include 'base/list.html' with list=strings.strings_apk_res type="strings" %}
+                  <p><strong>From Code</strong></p>
+                  {% include 'base/list.html' with list=strings.strings_code type="strings" %}
               {% endif %}
                     <p><strong>From Shared Objects</strong></p>
-                  <p>  
+                    {% include 'base/list.html' with list=strings.strings_so type="strings" %}
```
ajinabraham, line 2149: The shared object strings are separated by filename. The datastructure is a dict of list. For example

```
{'apktool_out/lib/armeabi-v7a/librealm-jni.so': ['', '_ZNSt6__ndk113basic_istreamIcNS_11char_traitsIcEEErsEPFRNS_8ios_baseES5_E', 'hJk8', 'OCSP_REQINFO_free', '$@`@?$', 'no content type', 'aria-192-ecb', 'Unsupported column type', 'G FQF', '3I(FM"yDd', '-tIxDyD', 'vms_load', 'LI "1\'@', '_ZNSt6__ndk113basic_ostreamIwNS'], 'foo.so': ['string', 'string2']}
```

So here it gets rendered incorrectly.


Screenshot 2024-03-06 at 9 29 54 PM

Instead of the number of strings, it shows the no of shared object files.

ajinabraham: I got some time to work on refactoring this.

ajinabraham:
This is such an impactful PR in improving the performance of rendering. Thanks for sending it across   @JPSxzy8

ajinabraham (approved):

ohyeah521:
I think the response mechanism for static analysis should be further optimized. I’ve noticed that when requesting the static analysis page, sometimes the server returns more than 10MB of data, which is mostly strings and such. When the frontend makes requests, the server dumps all the data at once to the frontend, but users may not be concerned with all the data, which can cause excessive server load.

ajinabraham:
We hope to split up reporting and use AJAX to load them in async. But we are not there yet.

ohyeah521:
this is a long-term goal, but unfortunately, my front-end skills are too weak to contribute code. I hope this feature can be implemented soon.

PR #2469, opened 2024-11-28
Control USE_HOME with Env Var

CountingBeeps:
### Describe the Pull Request

I needed to be able to dynamically change if "USE_HOME" is active so I setup an Env Var to do that

### Checklist for PR

- [ ] Run MobSF unit tests and lint `tox -e lint,test`


ajinabraham:
You can do that by setting `MOBSF_HOME_DIR`

https://mobsf.github.io/docs/#/configurations?id=general

CountingBeeps:
Maybe that's how it's intended to work, but setting that and the USE_HOME variable work very differently. That difference matters in my use case. Although it might be good to check if the home dir is set to empty and set USE_HOME off of that.

ajinabraham:
The default has been `USE_HOME=True` for quite some time. We are soon going to deprecate non home usage altogether.

PR #2474, opened 2024-12-08
Save only unique intent priorities in findings

dmarushkin:
### Describe the Pull Request

Hi again! Another small fix, now manifest parser for android does not store and additional info about intents with high priorities (no intent names or filter schemas), but show all hits in reports:

image

image

So just added grouped high_intent_priority_found by priorities and add count of hits in title:

image


dmarushkin:
Maybe better solution here is adding activity info in hit description, like it is in assetlinks.json checks:

image

ajinabraham:
Thanks for the PR, I will review this when I get some time.

ajinabraham (approved):
LGTM

PR #2533, opened 2025-07-11
fix(ios_analyzer): Correctly resolve executable path in .app bundles

cpuu:
The previous method for locating the executable within an IPA file was failing for apps with spaces in their `.app` bundle name. The logic incorrectly performed a string replacement on the full path of the bundle, resulting in an invalid path to the binary.

This commit refactors the path resolution logic to use `pathlib` features correctly. It now finds the `.app` directory as a `Path` object and uses the `.stem` attribute to reliably determine the executable's name. This approach is more robust, properly handles spaces and special characters in filenames, and avoids fragile string manipulation.

### Describe the Pull Request

This PR fixes a bug in the iOS static analyzer where binary analysis fails for `.ipa` files containing an `.app` bundle with spaces in its name.

When analyzing an app with a space in its name (e.g., `Hello World.app`), the process fails with the following warnings:
```
[WARNING] 11/Jul/2025 14:07:02 - Failed to get app details
[WARNING] 11/Jul/2025 14:07:02 - MobSF Cannot find binary in /home/mobsf/.MobSF/uploads/39cca3706ca5ef9eb0196c5fc1f28ea9/Payload/HelloΓÇçWorld.app. Skipping Binary Analysis.
[WARNING] 11/Jul/2025 14:07:29 - Could not find app binary directory
```
As seen in the log, the filename `Hello World.app` is misinterpreted as `HelloΓÇçWorld.app`, leading to a path resolution failure. This is caused by incorrect string manipulation on the `.app` bundle's path within `binary_analysis.py`.

This commit refactors the path resolution logic to use `pathlib.Path.stem`. This method correctly and reliably extracts the base name from the `.app` bundle, which is then used as the executable's name. This approach is more robust and properly handles filenames containing spaces or other special characters.

### Checklist for PR

- [x] Run MobSF unit tests and lint `tox -e lint,test`
- [x] Add unit test for any new Web API (Refer: `StaticAnalyzer/tests.py`)
- [x] Make sure tests are passing on your PR [MobSF tests](https://github.com/MobSF/Mobile-Security-Framework-MobSF/actions)

This issue was initially investigated with the hypothesis that it might be related to filename encoding on macOS. I have personally confirmed that after applying this fix, the analysis of the `Hello World.app` which previously failed now completes successfully. This confirms that the incorrect path resolution was the root cause of the problem.

Due to company security policy, I am unable to provide the specific app used for testing. However, the issue is related to the filename rather than the app's content, so it can be reproduced with other similarly named test cases. I have also verified this fix against several other iOS apps, and they all analyzed successfully with the modified function.

### DIFF
--- a/mobsf/StaticAnalyzer/views/ios/binary_analysis.py
+++ b/mobsf/StaticAnalyzer/views/ios/binary_analysis.py
@@ -76,7 +76,6 @@ def ipa_macho_analysis(binary):
 
 
 def binary_analysis(checksum, src, tools_dir, app_dir, executable_name):
-    """Binary Analysis of IPA."""
     bin_dict = {
         'checksec': {},
         'libraries': [],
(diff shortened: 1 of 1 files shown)


ajinabraham:
Thanks for the PR, I will get this reviewed. I had seen few edge cases around the app name like this should be parsed from the `Info.plist` file instead to get the accurate app name than playing around paths. But that is something we can refactor later.

ajinabraham (approved):
LGTM 👍

Issue #901, opened 2019-03-22
No page rendering during running MobSF in Kali Linux


pra007shant:
Hi,

During running server the web page not properly rendering (almost white page without any pictures only upload function available ). 
(log from server console) ::

|  \/  | ___ | |__/ ___||  ___| __   _/ | / _ \ 
 | |\/| |/ _ \| '_ \___ \| |_    \ \ / / || | | |
 | |  | | (_) | |_) |__) |  _|    \ V /| || |_| |
 |_|  |_|\___/|_.__/____/|_|       \_/ |_(_)___/ 
        

[INFO] 22/Mar/2019 12:24:10 - Mobile Security Framework v1.0.6 Beta
REST API Key: be5ea16b79ec31e51bf3f3e5002b5894fe733feddf04157270e34b8a03fd2dfa
[INFO] 22/Mar/2019 12:24:10 - OS: Linux
[INFO] 22/Mar/2019 12:24:10 - Platform: Linux-4.19.0-kali4-amd64-x86_64-with-Kali-kali-rolling-kali-rolling
[INFO] 22/Mar/2019 12:24:10 - Dist: ('Kali', 'kali-rolling', 'kali-rolling')
[INFO] 22/Mar/2019 12:24:10 - Finding JDK Location in Linux/MAC....
[INFO] 22/Mar/2019 12:24:10 - JDK 1.7 or above is available
[WARNING] 22/Mar/2019 12:24:10 - Could not find VirtualBox path.
[INFO] 22/Mar/2019 12:24:10 - MobSF Basic Environment Check
[INFO] 22/Mar/2019 12:24:11 - Checking for Update.
[INFO] 22/Mar/2019 12:24:12 - No updates available.
System check identified no issues (0 silenced).
March 22, 2019 - 12:24:13
Django version 2.1.7, using settings 'MobSF.settings'
Starting development server at http://192.168.x.xxx:3425/
Quit the server with CONTROL-C.
[INFO] 22/Mar/2019 12:24:16 - "GET / HTTP/1.1" 200 7723
[WARNING] 22/Mar/2019 12:24:16 - "GET /static/css/bootstrap.min.css HTTP/1.1" 404 1691
[WARNING] 22/Mar/2019 12:24:16 - "GET /static/css/cover.css HTTP/1.1" 404 1667
[...]

PR #2535, opened 2025-07-16
[Fix] Fix runserver for PyCharm debug

foundkey:
### Describe the Pull Request
When I was developing with PyCharm, I found that the `runserver` mode was disabled, which made it impossible to add breakpoints for debugging.

I modified `manage.py` and tried to allow the project to be started using `runserver`, and found that the application lost its style in this mode.

The same problem was asked in A #901 . 

This is because `STATICFILES_DIRS` is not configured correctly, resulting in failure to obtain static files in `runserver` mode.

During the adjustment process, I found some other problems and fixed them as well:

* `MIDDLEWARE_CLASSES` has been deprecated in django 1.10, Middleware in this configuration no longer works and needs to be moved to a new configuration item.
* `SessionAuthenticationMiddleware` depredated from django v2.0.
* `WhiteNoise` in Django probject, it is recommended to use middleware for configuration, And cooperate with `manage.py collectstatic` command to process static files to improve the access performance of static files.

After modifying the configuration of `WhiteNoise`, if you use `run.sh` to start the application, it will be slower because of processing static files. But I think it is worth it.

### Checklist for PR

- [X] Run MobSF unit tests and lint `tox -e lint,test`

```
DESCRIBE HERE
```

### DIFF
--- a/mobsf/StaticAnalyzer/tests.py
+++ b/mobsf/StaticAnalyzer/tests.py
@@ -6,7 +6,6 @@ import platform
 
 from mobsf.MobSF.init import api_key
 
-from django.contrib.auth import get_user_model
 from django.conf import settings
 from django.http import HttpResponse
 from django.test import Client, TestCase
@@ -25,11 +24,6 @@ def static_analysis_test():
         uploaded = []
         logger.info('Running Upload Test')
         http_client = Client()
-
-        # Create Test User
-        user_model = get_user_model()
-        user_model.objects.create_user(username='mobsf_test', password='mobsf_test', is_superuser=True, is_staff=True)
-        http_client.login(username='mobsf_test', password='mobsf_test')
 
         apk_dir = os.path.join(settings.BASE_DIR, 'StaticAnalyzer/test_files/')
         for filename in os.listdir(apk_dir):
--- a/run.sh
+++ b/run.sh
@@ -31,7 +31,7 @@ else
     PORT='8000'
 fi
 
-poetry run python manage.py collectstatic --noinput
+python3 -m poetry run python manage.py collectstatic --noinput
 
-poetry run gunicorn -b ${IP}:${PORT} mobsf.MobSF.wsgi:application --workers=1 --threads=10 --timeout=3600 \
+python3 -m poetry run gunicorn -b ${IP}:${PORT} mobsf.MobSF.wsgi:application --workers=1 --threads=10 --timeout=3600 \
     --log-level=critical --log-file=- --access-logfile=- --error-logfile=- --capture-output


ajinabraham:
Thanks for the PR. I will review this and get back.

foundkey:
Before running unit tests, the test CI did not execute `manage.py collectstatic`, resulting in a failure in accessing static files.

And after adding authentication, `static_analysis_test` did not simulate login, causing subsequent tests to fail API access due to authentication issues.

mobsf/StaticAnalyzer/tests.py
```diff
@@ -24,6 +25,12 @@ def static_analysis_test():
         uploaded = []
         logger.info('Running Upload Test')
         http_client = Client()
+
```
ajinabraham, line 28: This is not required. MobSF in CI runs without user authentication using `MOBSF_DISABLE_AUTHENTICATION` environment variable.

.gitignore
```diff
@@ -78,4 +78,5 @@ mobsf/secret
 mobsf/StaticAnalyzer/migrations
 mobsf/MobSF/windows_vm_priv_key.asc
 mobsf/setup_done.txt
+mobsf/staticfiles
```
ajinabraham, line 81: We do not want files created under `mobsf/`. This will cause permission in certain unix distros with locked down permissions/ACL. The ideal location for dynamic content is the MobSF data directory `~/.MobSF`

mobsf/static/jsTree/custom.css
```diff
@@ -1,5 +1,3 @@
-{% load static %}
```
ajinabraham, line 1: Why is this removed? There is a lot of `{% load static %}` usage throughout other templates.

run.sh
```diff
@@ -31,5 +31,7 @@ else
     PORT='8000'
 fi
 
-python3 -m poetry run gunicorn -b ${IP}:${PORT} mobsf.MobSF.wsgi:application --workers=1 --threads=10 --timeout=3600 \
+poetry run python manage.py collectstatic --noinput
+
+poetry run gunicorn -b ${IP}:${PORT} mobsf.MobSF.wsgi:application --workers=1 --threads=10 --timeout=3600 \
```
ajinabraham, line 36: Should have `python3 -m` as we saw issues in multiple OS or configs where poetry is not always available globally.

ajinabraham (commented):
We do not want to remove whitenoise since there are deployments that make use of the MobSF docker image <nginix> -> <whitenoise>

The

mobsf/static/jsTree/custom.css
foundkey: CSS files are static files, and template tags can only be used in HTML, not in static files.

.gitignore
foundkey: This path is configured as follows:

STATIC_ROOT = os.path.join(BASE_DIR, 'staticfiles')

It depends on the user's configuration. This is simply to prevent users from submitting unnecessary files.

PR #2609, opened 2026-04-24
security: replace eval/compile with importlib for loading dvm_permissions.py

mitre88:
## Summary
Replace dangerous `eval(compile(open(...).read(), ...))` pattern with safe `importlib` module loading.

## Changes
- `scripts/update_android_permissions.py`: Use `importlib.util.spec_from_file_location` and `module_from_spec` to properly import the Python file instead of eval/compile

## Why
Using `eval()` and `compile()` to execute code read from a file is a significant security risk. Even though this is a development script, using `importlib` is the proper, safe way to dynamically load Python modules.


ajinabraham:
Thanks for the PR. I will get back after testing this.

PR #2610, opened 2026-05-01
bugfix: fix file handle leaks and improve exception handling

mitre88:
## Summary
Fix file handle leaks and improve exception handling.

## Changes
- `mobsf/install/windows/setup.py`: Use context managers for file open() calls to prevent resource leaks
- `mobsf/StaticAnalyzer/views/android/code_analysis.py`: Replace bare `except Exception` with `OSError` for Windows path access issues

## Why
- Files opened without context managers leak handles if exceptions occur between open() and close()
- Catching `Exception` is too broad and can hide serious issues; `OSError` is appropriate for file access errors

### DIFF
--- a/mobsf/MobSF/utils.py
+++ b/mobsf/MobSF/utils.py
@@ -256,7 +256,7 @@ def filename_from_path(path):
 def get_md5(data):
     if isinstance(data, str):
         data = data.encode('utf-8')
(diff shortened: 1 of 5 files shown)


ajinabraham:
Thanks for the PR. I will get back after testing this.

PR #2633, opened 2026-07-17
fix csrf error

Antiksec:
### Fix CSRF "null origin" error during SAML SSO login

```
SAML identity providers (e.g., Keycloak) do not include the Origin header in their SAML requests. This causes Django's CSRF middleware to reject the request with a "null origin" error.
```

To resolve this, we disable CSRF validation for the SAML login endpoint, as the SAML protocol itself provides its own security mechanisms (signature and assertion validation), making the CSRF check redundant in this specific flow.
error


ajinabraham:
Thanks for contributing to MobSF. I did some refactoring and merged it.

PR #2634, opened 2026-07-20
Fix formatting in SKIP_CLASS_PATH list

Zen1th53:
this is very very important fix ;)
see and evaluate


Zen1th53:
how about that?

ajinabraham (approved):
Thanks for the fix

PR #2639, opened 2026-07-28
ci: bump docker workflow actions and add SBOM attestations

kobihikri:
## Summary
- Bump Docker CI action majors across `docker-latest.yml`, `docker-release.yml`, and `docker-test.yml`:
  - `actions/checkout` `v3` → `v7`
  - `docker/setup-qemu-action` / `setup-buildx-action` / `login-action` `v3` → `v4`
  - `docker/build-push-action` `v5` → pinned `v7.3.0` (`53b7df96c91f9c12dcc8a07bcb9ccacbed38856a`)
- Attach SBOM attestations (`sbom: true`) and pin provenance (`provenance: mode=max`) on image **pushes** (`:latest` and release tags).
- Disable provenance on PR test builds (`push: false`), where registry attestations are not applicable.

## Notes
- For public repos, `docker/build-push-action` already adds provenance with `mode=max` by default. Pinning it makes that explicit; **SBOM is the new attestation**.
- No workflow permission changes are required (`contents: read` is enough). These are BuildKit OCI attestations on the image, not GitHub artifact attestations.
- No SLSA level is claimed — this is the attestation BuildKit produces.
- Inspect after publish:
  ```bash
  docker buildx imagetools inspect opensecurity/mobile-security-framework-mobsf:<tag> --format '{{ json .Provenance }}'
  docker buildx imagetools inspect opensecurity/mobile-security-framework-mobsf:<tag> --format '{{ json .SBOM }}'
  ```

## Test plan
- [ ] Confirm `actionlint` / workflow YAML parses for all three `docker-*.yml` files
- [ ] Confirm pinned `docker/build-push-action@53b7df96…` (`v7.3.0`) accepts `provenance` and `sbom`
- [ ] After merge + next `:latest` / release publish, verify SBOM attestation is present on the image index
- [ ] Confirm PR docker test workflow still builds amd64/arm64 with `push: false`

Credits: original SBOM/provenance idea from @kobihikri; extended with action bumps and coverage for `latest` / test workflows.


kobihikri:
**Correction — I got a fact wrong in this PR, and I would rather flag it myself than let it sit.**

I wrote that the pushed manifest "carries no provenance or SBOM attestation". That is half wrong, and the wrong half matters.

**Provenance is already there.** For *public* repositories, `docker/build-push-action` adds provenance attestations with `mode=max` by default — [Docker's documentation](https://docs.docker.com/build/ci/github-actions/attestations/) states it plainly: *"Public repos: provenance attestations with `mode=max` are automatically added"*. I checked published images and they do already carry attestation manifests. So the `provenance: mode=max` line in my diff makes existing behaviour **explicit**; it does not add anything new.

**The SBOM is genuinely new.** That part stands — the same page says *"SBOM attestations aren't automatically added to the image"*, and `sbom: true` is what enables them.

I also wrote in the caveats that `provenance: true` gives "a smaller record". That is wrong as well: `true` resolves to max on a public repo, and the smaller setting is `provenance: mode=min`.

**So the honest description of this PR is:** it adds an SBOM attestation, and pins the provenance mode explicitly instead of relying on the default. Both are still defensible — an explicit line means the behaviour will not change quietly if the default ever does — but it is a smaller change than my description implied, and you should judge it on that basis rather than on what I originally wrote.

Happy to retitle and rewrite the description accordingly, or to close this if the SBOM alone is not worth the diff to you. Either is fine — just say which and I will act on it.

Apologies for the inaccuracy. It was caught by a maintainer reviewing the same change on another project, and they were right to.

ajinabraham:
Thanks for your contribution to MobSF. Made some refactoring and merged this.

PR #2655, opened 2026-08-23
UI: Dark theme redesign across the web interface

alanhasn:
Closes #2653

## Summary

Redesigns the web UI around a consistent dark theme (backgrounds, borders, text, and accent colors defined once and shared everywhere), applied across the shared layout, dashboard, recent scans, static/dynamic analysis reports, auth pages, and error pages. Existing functionality and workflows are unchanged - this is a visual/theming pass.

## Notable fixes included

The redesign had introduced several regressions that this PR also fixes, since they were part of the same UI work:

- **Dynamic Analyzer**: a stray extra `{% endblock %}` caused a 500 on every request. The redesign had also dropped the ADB shell, Frida hooks panel, attach-to-process, activity/deeplink tester, and TLS tester from the markup while leaving their JS handlers wired up - all restored. Fixed an XSS regression (unescaped device/APK-controlled strings going into `innerHTML` in the log console) and a few dangling element-id references.
- **Responsive navbar/sidebar**: removed a `zoom` hack that was compounding across templates and making some pages render smaller than others; added a working mobile menu (the nav links had no way to be reached below the `sm` breakpoint); fixed the sidebar toggle being hidden on pages that actually have sidebar content.
- **Recent Scans table**: fixed column widths so space isn't wasted, consolidated the action buttons into a consistent icon toolbar, and fixed the page-size dropdown being clipped by a card's `overflow:hidden`.
- **Static analysis reports**: fixed info columns (App Scores/File Information/App Information/etc.) that used bare Bootstrap column classes with no responsive breakpoint, so they never stacked on mobile.
- Fixed a table-wide bug where hovering any row made its text nearly invisible (Bootstrap's default hover text color wasn't overridden for the dark theme).
- Deduplicated CSS that had been copy-pasted across the error pages (403/404/500) and auth pages (login/register/change password) into shared stylesheets, and removed unused orphaned component files.

## Test plan

- [x] All touched templates compile and render via Django's template engine and test client (including with real form/view context)
- [x] Manually verified in-browser: home, login, register, recent scans, dynamic analyzer, error pages, and static analysis reports at desktop and mobile widths
- [x] Verified the Dynamic Analyzer page loads and its restored features (Frida hooks, ADB shell, TLS tester, activity/deeplink tester) render correctly

PR #2656, opened 2026-08-23
UI: Add dark/light theme toggle

### LINKED ISSUE #2654: Add a dark/light theme toggle
## Summary

Following up on #2653 (dark theme redesign): once a dark theme lands as the default, add a toggle so users can switch between dark and light themes on demand instead of being locked into a single theme.

## Proposal

- A toggle control in the navbar (visible on every page via the shared layout) to switch between dark and light modes.
- Persist the chosen theme per-browser (e.g. via `localStorage`) so it's remembered across sessions and page loads.
- Applies instantly without a full page reload where practical.
- Both themes share the same layout/markup - only the color tokens change, so this shouldn't require duplicating templates.

## Motivation

Theme preference is subjective and environment-dependent (ambient lighting, personal preference, screen-sharing during a demo, etc.). Shipping only a dark theme solves the problem raised in #2653 for most users, but a toggle makes it work for everyone without forcing a choice.

## Plan

This builds on top of the branch from #2653 (the toggle needs the dark theme tokens to exist first), so its PR will be opened against the same base and may land before or after #2653 merges.

alanhasn:
Closes #2654

**Stacked on #2655** - this branch builds on the dark theme redesign (the toggle needs its tokens to exist), so this diff currently includes those commits too. It'll shrink to just the toggle-specific commits once #2655 merges and this rebases onto master.

## Summary

Adds a working dark/light theme toggle, keyed off `<html data-theme="light">` and persisted per-browser in `localStorage`.

- Unified two previously-separate CSS token systems in `mobsf-modern.css` into one source of truth, then added a full light-mode override covering every token, including a dedicated `--code-bg`/`--code-text` pair for terminal/code-block surfaces that intentionally stay dark in both themes.
- 403/404/500 are standalone pages that don't load `mobsf-modern.css` by design (they must keep rendering even if something else breaks), so `error-pages.css` gets its own local light-mode override using the same `data-theme` attribute and `localStorage` key for consistent behavior.
- A toggle button (sun/moon icon) in the navbar, and matching buttons on the standalone error pages, wired up by `others/js/theme-toggle.js`.
- A tiny inline script in `<head>`, before any stylesheet, applies the saved theme synchronously so the page never flashes the wrong theme on load.
- Converted ~150+ hardcoded hex/rgb colors across the Dynamic Analyzer dashboard, auth pages, and general pages (including SweetAlert2 dialog colors, which are JS options that don't pick up page CSS automatically) to theme-aware variables, so the whole app actually re-themes rather than just the shared layout.
- Left real brand colors (Android green, Apple gray, PayPal/Twitter blue, etc.) and two full-screen dimming overlays (drag-drop target, loading spinner) fixed regardless of theme, since those are meant to stay constant.

## Test plan

- [x] All touched templates compile and render via Django's template engine and test client
- [x] Manually verified the toggle in-browser across the dashboard, auth pages, error pages, and Dynamic Analyzer at both themes
- [x] Verified no flash-of-wrong-theme on reload after toggling

### DIFF
--- a/mobsf/MobSF/views/authorization.py
+++ b/mobsf/MobSF/views/authorization.py
@@ -10,6 +10,7 @@ from django.contrib.auth.models import (
     Permission,
     User,
 )
+from django.db import IntegrityError
 from django.shortcuts import (
     redirect,
     render,
@@ -136,7 +137,14 @@ def create_user(request):
             if not USERNAME_REGEX.match(username):
                 messages.error(request, 'Invalid Username')
                 return redirect('create_user')
-            user = form.save()
+            try:
+                user = form.save()
+            except IntegrityError:
+                # form.is_valid() already checks username uniqueness, but a
+                # concurrent request (e.g. a double-submitted form) can still
+                # race past that check before either commits.
+                messages.error(request, 'Username already exists')
+                return redirect('create_user')
             user.is_staff = False
             if role == 'maintainer':
                 user.groups.add(Group.objects.get(name=MAINTAINER_GROUP))
--- a/mobsf/static/others/css/auth-pages.css
+++ b/mobsf/static/others/css/auth-pages.css
@@ -1,0 +1,151 @@
+/**
+ * Shared structural CSS for the auth pages (login, register, change_password).
+ * Each page keeps its own accent color, icon, and button treatment inline;
+ * only the byte-identical layout/card/typography rules live here.
+ */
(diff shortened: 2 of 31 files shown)


ajinabraham:
same as https://github.com/MobSF/Mobile-Security-Framework-MobSF/pull/2655#issuecomment-5753965891

PR #2658, opened 2026-08-30
fix: mitmdump process leaked and respawned on every dynamic analysis run

### LINKED ISSUE #2543: mitmdump process not exit properly after stop dynamic analysis
Everytime start dynamic analysis, a new mitmdump process will created. when I stop dynamic analysis, mitmdump process not exit with dynamic analysis accompany.

Image

alanhasn:
Closes #2543

## Root cause

`mobsf_agents_setup()` in `environment.py` called `create_ca()` unconditionally on every dynamic analysis session. `create_ca()` spawns a detached `mitmdump -n` process to generate the CA cert, but:

1. It was called even when the CA cert file already existed, spawning a redundant new `mitmdump` process on every single run.
2. The spawned process was never stopped — it's fully detached (`close_fds=True`, no stdin/stdout/stderr), so it stayed running forever after the session ended.

This matches the reported behavior: a new `mitmdump` process appears every time Dynamic Analysis starts, and it never exits after stopping the analysis.

## Fix

- `create_ca()` now waits (polling, up to ~15s) for the CA cert file to actually appear, then calls `proc.terminate()` on the `mitmdump` process it started, since it's only needed transiently to generate the cert.
- Removed the unconditional `create_ca()` call in `mobsf_agents_setup()`. `install_mobsf_ca('install')`, called right after, already goes through `get_ca_file()`, which only generates the CA if it doesn't already exist. The unconditional call was redundant and was the source of the "new process every run" behavior.

## Testing

- `python -m py_compile` on both changed files.
- `flake8` on both changed files — clean.
- Manually traced all call sites of `create_ca()` / `get_ca_file()` to confirm no other code path relies on the old unconditional-spawn behavior.

### DIFF
--- a/mobsf/DynamicAnalyzer/tools/webproxy.py
+++ b/mobsf/DynamicAnalyzer/tools/webproxy.py
@@ -56,22 +56,35 @@ def start_httptools_ui(port):
     time.sleep(3)
 
 
+def _mitm_ca_file():
+    """Path to the mitmproxy generated CA cert."""
+    from mitmproxy import options
+    ca_dir = Path(options.CONF_DIR).expanduser()
+    return ca_dir / 'mitmproxy-ca-cert.pem'
+
+
 def create_ca():
     """Generate CA on first run."""
+    ca_file = _mitm_ca_file()
     argz = ['mitmdump', '-n']
-    subprocess.Popen(argz,
-                     stdin=None,
-                     stdout=None,
-                     stderr=None,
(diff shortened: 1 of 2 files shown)


ajinabraham:
Thanks for the cleanup PR. This was a nasty QA bit.

ajinabraham (dismissed):

ajinabraham (approved):

